PR #1, opened 2020-04-12
Project setup

tommylettieri:


package.json
```diff
@@ -0,0 +1,38 @@
...
+    "@types/bluebird": "^3.5.30",
+    "@types/express": "^4.17.6",
+    "@types/node": "^13.11.1",
+    "@types/validator": "^13.0.0",
+    "sequelize-cli": "^5.5.1",
+    "typescript": "^3.8.3"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "npm run build && node ts-built/index.js",
```
gbhojraj, line 24: I'm assuming this currently requires a manual restart, and we'll add nodemon in later?

src/database/index.ts
```diff
@@ -0,0 +1,20 @@
+
+import appSequelize from './app-sequelize';
+
+import user from './models/user';
+
+(async () => {
```
gbhojraj, line 6: Does this need to eventually be `await`ed?

package.json
tommylettieri: Yup

tommylettieri: I didn't have a chance to get nodemon, with typescript working, especially getting it to work with stack traces (source maps) currently we do this with ts-node

PR #2, opened 2020-05-02
Project setup

tommylettieri:
Added nodemon for quicker auto restarts and quicker deployment
Added response formatting to limiter
Added failure case to account verification
Added linter
Fixed all linting issues
Fixed register call to not return sensitive cases
Added awaits for startup to avoid errors
Added permission model
Added permission logic on register

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -9,7 +9,7 @@ import permission from './models/permission';
 export const sync = async (): Promise<void> => {
     try {
         await appSequelize.authenticate();
-        await appSequelize.sync()
+        await appSequelize.sync();
     } catch (e) {
         logger.error('Could not init sequelize', e)
     }
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -4,6 +4,7 @@ import appSequelize from './app-sequelize';
 import user from './models/user';
 import university from './models/university';
 import session from './models/session';
+import permission from './models/permission';
 
 export const sync = async (): Promise<void> => {
     try {
@@ -18,6 +19,7 @@ const database = {
     user,
     university,
     session,
+    permission,
     appSequelize
 }
 
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -5,14 +5,15 @@ import user from './models/user';
 import university from './models/university';
 import session from './models/session';
 
-(async (): Promise<void> => {
+export const sync = async (): Promise<void> => {
     try {
-        appSequelize.authenticate();
-        appSequelize.sync()
+        await appSequelize.authenticate();
+        await appSequelize.sync()
     } catch (e) {
         logger.error('Could not init sequelize', e)
     }
-})();
+};
+
 const database = {
     user,
     university,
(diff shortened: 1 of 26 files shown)


src/database/index.ts
```diff
@@ -4,19 +4,22 @@ import appSequelize from './app-sequelize';
...
 import session from './models/session';
+import permission from './models/permission';
 
-(async () => {
+export const sync = async (): Promise<void> => {
     try {
-        appSequelize.authenticate();
-        appSequelize.sync()    
+        await appSequelize.authenticate();
+        await appSequelize.sync()
```
gbhojraj, line 12:
```suggestion
        await appSequelize.sync();
```

src/features/users/user-controller.ts
```diff
@@ -115,24 +116,33 @@ class UserController {
         if (universities.length > 1) {
             logger.error(`Multiple universities found ${universities.length}`);
         }
-        university = universities[0];
+        const university = universities[0];
 
 
         if (university.student_email_domain === emailDomain) {
-            logger.info('User is student');
+            // Database field
```
gbhojraj, line 123: Logger statements could still be useful, no?

src/features/users/user-route.ts
```diff
@@ -15,8 +15,10 @@ import AlreadyExistsError from "../../exceptions/already-exists-error";
 router.post('/login',
     validate(loginValidation),
     passport.authenticate('local'),
-    asyncHandler(async (req: any, res: Response, next: NextFunction) => {
-        const newSession = req.session.passport.user;
+    asyncHandler(async (req: Request, res: Response, next: NextFunction) => {
+        // TODO fix Request object for session
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
gbhojraj, line 20: Do you still need this disable-line?

src/middleware/joi-validator.ts
```diff
@@ -1,26 +1,33 @@
...
 
     // Using Request from express I get typescript errors
-    return function validateRequest(req: any, res: Response, next: any) {
-        let toValidate: any = {};
+    return function validateRequest(req: Request, res: Response, next: NextFunction): void {
+        // This is a generic callback which can be used for any object, thus this can be any as well
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const toValidate: any = {};
         if (!schema) {
             next();
```
gbhojraj, line 15: Should this be `return`ed?

src/features/users/user-controller.ts
tommylettieri: Seems like bloat when it is in the response

src/middleware/joi-validator.ts
tommylettieri: Good call, doesn't break but it probably can cause an issue

src/features/users/user-route.ts
tommylettieri: Looks like it from here, but not from the source, maybe this is stale

PR #6, opened 2020-05-09
Models branch

tommylettieri:


src/database/models/session.ts
```diff
@@ -47,7 +47,7 @@ Session.init({
 });
 
 Session.belongsTo(User, {
-    foreignKey: 'user_id',
+    foreignKey: 'userId',
```
gbhojraj, line 50: Is this supposed to be the foreignKey in the database or on the class?

src/features/users/user-controller.ts
```diff
@@ -154,7 +142,7 @@ class UserController {
             await emailHelper.sendEmail({
                 content: `Hello,
 
-                Please verify your account by clicking this url: ${baseUrl}/users/verify?verify_token=${newUser.verify_token}
+                Please verify your account by clicking this url: ${baseUrl}/users/verify?verify_token=${newUser.verifyToken}
                 `,
                 email: newUser.email,
                 subject: 'Please veryify account'
```
gbhojraj, line 148: Spelling error (should be fixed in this PR).
```suggestion
                subject: 'Please verify your Rederly account'
```

gbhojraj:
I'm also not sure if it belongs in this PR, but can you change username to First/Last name? @tommylettieri

src/features/users/user-controller.ts
```diff
@@ -154,7 +142,7 @@ class UserController {
             await emailHelper.sendEmail({
                 content: `Hello,
 
-                Please verify your account by clicking this url: ${baseUrl}/users/verify?verify_token=${newUser.verify_token}
+                Please verify your account by clicking this url: ${baseUrl}/users/verify?verify_token=${newUser.verifyToken}
```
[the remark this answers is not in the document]
gbhojraj: Actually, I'll let you mess with the configuration as you want. The appropriate path should be:
```
${frontendUrl}/verify/${newUser.verify_token}
```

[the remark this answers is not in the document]
tommylettieri: the front end url should be the same as the base url due to the proxy, I'll get rid of /users and fix the typeo

src/database/models/session.ts
tommylettieri: Sequelize accepts both, but I would rather use our field since we might change database field name but we won't change the key we use in the code, fixed for other comment

PR #9, opened 2020-05-17
Emails

tommylettieri:


src/features/users/user-controller.ts
```diff
@@ -37,18 +48,51 @@ class UserController {
         })
     }
 
-    list(): Bluebird<User[]> {
+    list(listOptions?: ListOptions<ListUserFilter>): Bluebird<User[]> {
+        // Dynamic sequelize where object
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const where: any = {};
+        if (listOptions) {
```
gbhojraj, line 55: Can you use the null-safe accessor?

```
if (listOptions?.filters?.userIds)```

tommylettieri: This is more of a micro optimization

in this case if you have 1 filter it will do 3 checks
if you had 3 filters there would be 5 checks

with null-safe accessor in this case 1 filter would be the same
but if you had 3 filters there would be 9 checks

likewise if we have page options we can also include that here to be more efficient

PR #17, opened 2020-06-06
Grades

tommylettieri:


src/features/users/user-controller.ts
```diff
@@ -9,12 +10,18 @@ import Session from '../../database/models/session';
 import moment = require('moment');
 import configurations from '../../configurations';
 import NoAssociatedUniversityError from '../../exceptions/no-associated-university-error';
-import { UniqueConstraintError } from 'sequelize';
+import { UniqueConstraintError, Op } from 'sequelize';
```
gbhojraj, line 13: What is `Op`? I don't see it being used here.

tommylettieri: Good catch, I removed the code that used it, i'll remove that in the future

PR #57, opened 2020-07-24
Jm/dev branch

Jhuanderson-Rederly:
Change the verify endpoint URL
Down the line, we would want to route the backend url to demo.rederly.com/backend-api/etc..
and redirect them straight to rederly if status = 200 ok

### DIFF
--- a/src/features/users/user-controller.ts
+++ b/src/features/users/user-controller.ts
@@ -334,7 +334,7 @@ class UserController {
             await emailHelper.sendEmail({
                 content: `Hello,
 
-                Please verify your account by clicking this url: ${baseUrl}/verify/${newUser.verifyToken}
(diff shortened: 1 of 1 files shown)


src/features/users/user-controller.ts
```diff
@@ -334,7 +334,7 @@ class UserController {
             await emailHelper.sendEmail({
                 content: `Hello,
 
-                Please verify your account by clicking this url: ${baseUrl}/verify/${newUser.verifyToken}
+                Please verify your account by clicking this url: ${baseUrl}/backend-api/users/verify?verifyToken=${newUser.verifyToken}
```
red-tt-bear, line 337: This is the incorrect url, this will point to the backend but we need the front end to make the request so that they can render the page and present an error message if required

doombot-exe: also, I assumed an incorrect `${baseUrl}` in my previous comment...

PR #59, opened 2020-07-24
Change problem rendering template

doombot-exe:
incomplete addressing of rederly/frontend#59

this changes the buttons that appear - but not the permissions for scaffold


src/features/courses/course-route.ts
```diff
@@ -235,7 +235,7 @@ router.post('/question/:id',
         proxyReqPathResolver: (req) => {
             return `/rendered?${qs.stringify({
                 format: 'json',
-                template: 'simple',
+                template: 'single',
```
red-tt-bear, line 238: This is good but I think we also need this on get question (in the file `renderer-helper.ts`

Issue #32, opened 2020-06-21
Database mapping fixes


red-tt-bear:
With sequelize we are able to change field names and not impact the code.  With this in mind we have shortened field name and made them conform to standard javascript standards

PR #65, opened 2020-07-25
Issue #32 Fixed unique constraints, Added owned university relationsh…

red-tt-bear:
…ip, Added textbooks to curriculum


Jhuanderson-Rederly:
I see you removed the unique constraint on the curriculum name and the course name. Is there a constraint on unique curriculum name per university & unique course name per university as well. Isn't that the route we were going for these two constraints? Correct me if I remembered wrong! =)

PR #86, opened 2020-08-09
Feature/add users filter to grades

PanopticaRising:


PanopticaRising:
We're going to move this functionality to the statistics call.

PR #87, opened 2020-08-13
Feature/red 41 students can see own grades like statistics

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -439,14 +439,15 @@ class CourseController {
...
+            unitId,
+            userId,
         } = options.where;
 
         const setFilterCount = [
             courseId,
             questionId,
             topicId,
-            unitId
+            unitId,
```
red-tt-bear, line 450: You didn't add userId here? should it be?  maybe a comment as to why not

red-tt-bear: If this is the case it is saying that you can't get all grades for a user for all courses which might be alright

PR #118, opened 2020-08-31
preserve sequelize jsonb and handle NUL bytes smrtly

doombot-exe:
I don't like stringify on JSON for the database -- this changes how we parse the renderer result, creating arrays in form_data wherever the original multipart-form/data included repeated keys. It will be much easier to reconstruct the student submission from this structure... and we keep the jsonb-enefits in pg


doombot-exe:
the renderer initially receives the form data with duplicated key(s)... it should be left to the renderer to make this known (without relying on NUL)

duplicated key(s) will be reported back by the renderer as arrays in JSON instead of NUL-delimited strings.

PR #134, opened 2020-09-15
Feature/red 9 grade override

red-tt-bear:


src/utilities/renderer-helper.ts
```diff
@@ -179,7 +179,7 @@ class RendererHelper {
     async getProblem({
         sourceFilePath,
         problemSource,
-        problemSeed,
+        problemSeed = 666,
```
PanopticaRising, line 182: Why is this defaulted to 666 here and defaulted to [here](https://github.com/rederly/backend/pull/134/files#diff-1fd04e20c23b5de82549b9541a8df404R878)?

src/features/courses/course-repository.ts
```diff
@@ -445,6 +446,25 @@ class CourseRepository {
...
+        try {
+            const updates = await StudentGrade.update(options.updates, {
+                where: {
+                    ...options.where,
+                    active: true
+                },
+                returning: true,
+            });
+            return {
+                updatedCount: updates[0],
```
PanopticaRising, line 460: Do we need to document the mapping of these array returns?

PanopticaRising (approved):

src/utilities/renderer-helper.ts
doombot-exe: renderer will always default a missing problemSeed to 666 regardless

red-tt-bear: I could remove it, probably makes sense, the doc had it labeled as required so I included it 😆

src/features/courses/course-repository.ts
red-tt-bear: Not sure what you're asking, but:
1. The response is typed
2. This is not actually an array, it's a tuple

src/utilities/renderer-helper.ts
PanopticaRising: I was just confused cause you changed the 666 above to null, so I was wondering why the inconsistency.

PR #146, opened 2020-09-24
Feature/red 109 exams fields

doombot-exe:
enable currentProblemState for problems that have been interacted with

### DIFF
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -849,6 +849,9 @@ class CourseController {
     }
 
     async getQuestion(options: GetQuestionOptions): Promise<GetQuestionResult> {
+        // grades/statistics may send workbookID => show problem with workbookID.form_data
+        // (not enrolled) problem page will send questionID without userID => show problem with no form_data
+        // (enrolled) will send questionID with userID => show problem with grades.currentProblemState
         const courseQuestion = await this.getQuestionRecord(options.questionId);
 
         if (_.isNil(courseQuestion)) {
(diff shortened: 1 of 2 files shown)


PanopticaRising (approved):

src/features/courses/course-controller.ts
```diff
@@ -875,20 +880,32 @@ class CourseController {
...
-        
 
-        if(_.isNil(workbook)) {
-            const workbooks = await studentGrade?.getWorkbooks({
-                limit: 1,
-                order: [ [ 'createdAt', 'DESC' ]]
-            });
-            workbook = workbooks?.[0] || null;
+        // if no workbookID, get the most recent workbook -- come back and delete this?
+        // if not enrolled, we don't even want to have workbooks
```
red-tt-bear, line 888: If student grade is null (not enrolled) student grade would be null
since it is null it wouldn't actually call get workbooks, instead it would return undefined
then workbook would be null

src/features/courses/course-controller.ts
```diff
@@ -875,20 +880,32 @@ class CourseController {
...
+        // GetProblemParameters requires undefined over null
+        let formData: {[key: string]: unknown} | undefined = studentGrade?.currentProblemState;
+        // at this point, we only have a workbook if a valid workbookID was provided
+        // set the formData to match the contents of the workbook
+        if(!_.isNil(workbook)) {
+            formData = workbook.submitted.form_data;
         }
 
-
+        // studentGrade is the source of truth
```
red-tt-bear, line 908: although in the future if a workbook is supplied it would be the source of truth
by the future I am talking about show me another or update randomization but you are looking at a previous attempt
so it would be:
```typescript
const randomSeed = workbook?.randomSeed ?? studentGrade?.randomSeed ?? null;
```

red-tt-bear (approved):
The comments I left were just thoughts based on the comments, pr is perfect!

PR #148, opened 2020-09-25
Feature/red 100 auditing

red-tt-bear:
This one is going to need some dev testing before merging


src/features/courses/course-repository.ts
```diff
@@ -532,9 +536,28 @@ class CourseRepository {
         }
     }
 
+    async createStudentGradeOverride(obj: Partial<StudentGradeOverride>): Promise<StudentGradeOverride> {
+        try {
+            return await StudentGradeOverride.create(obj);
+        } catch (e) {
+            // Check constraint errors if there are any
```
PanopticaRising, line 543: Does the Wrapped error check constraints itself?

src/features/courses/course-controller.ts
```diff
@@ -878,12 +885,40 @@ class CourseController {
...
             return resultantUpdates;
         });
     }
 
-    updateGrade(options: UpdateGradeOptions): Promise<UpdateResult<StudentGrade>> {
-        return courseRepository.updateGrade(options);
+    async updateGrade(options: UpdateGradeOptions): Promise<UpdateResult<StudentGrade>> {
+        return useDatabaseTransaction(async (): Promise<UpdateResult<StudentGrade>> => {
+            if (!_.isNil(options.updates.effectiveScore)) {
+                await courseRepository.createStudentGradeOverride({
```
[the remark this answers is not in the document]
red-tt-bear: Nope, the error should be bubbled up, if it can't record the audit then it shouldn't let the override through

src/features/courses/course-controller.ts
```diff
@@ -1050,6 +1087,227 @@ class CourseController {
...
+            studentGrade.latestAttempts = gradeResult.score;
+            studentGrade.numAttempts++;
+            if (gradeResult.gradingPolicy.isOnTime && !gradeResult.gradingPolicy.isLocked && gradeResult.gradingPolicy.isWithinAttemptLimit) {
+                studentGrade.numLegalAttempts++;
+            }
+            if (!gradeResult.gradingPolicy.isExpired && !gradeResult.gradingPolicy.isLocked && gradeResult.gradingPolicy.isWithinAttemptLimit) {
+                studentGrade.numExtendedAttempts++;
+            }
+
+            if (_.isNil(workbook)) {
```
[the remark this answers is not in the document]
red-tt-bear: The retro feature that this PR has in it

src/features/courses/course-controller.ts
```diff
@@ -1050,6 +1087,227 @@ class CourseController {
...
+            });
+            question = question ?? await studentGrade.getQuestion();
+            topic = topic ?? await question.getTopic();
+            const solutionDate = moment(topic.deadDate).add(Constants.Course.SHOW_SOLUTIONS_DELAY_IN_DAYS, 'days');
+
+            // reset student grade before submitting
+            studentGrade.bestScore = 0;
+            studentGrade.overallBestScore = 0;
+            studentGrade.partialCreditBestScore = 0;
+            // studentGrade.effectiveScore // Don't touch this
```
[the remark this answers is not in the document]
red-tt-bear: Sure I'll add that

src/features/courses/course-repository.ts
red-tt-bear: Nope, all it does is combine the stack of the error with the current stack
sequelize errors stacks never point back to our code which is why this is so helpful
This create doesn't take any user input that should break it, therefore if there is an issue it's an internal server error

src/features/courses/course-route.ts
```diff
@@ -371,14 +370,19 @@ router.put('/question/grade/:id',
     // This is to work around "extractMap" error
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     asyncHandler(async (req: RederlyExpressRequest<any, unknown, UpdateGradeRequest.body, UpdateGradeRequest.query>, _res: Response, next: NextFunction) => {
+        if (_.isNil(req.session)) {
```
PanopticaRising, line 373: Should we move this up to a middleware? It needs to be everywhere except login/public routes.

src/features/courses/course-types.ts
```diff
@@ -299,6 +307,33 @@ export interface GetCalculatedRendererParamsResponse {
     showSolutions: number;
 }
 
+export interface GradeOptions {
+    studentGrade: StudentGrade;
+    topic: CourseTopicContent;
+    question: CourseWWTopicQuestion;
+
+    solutionDate: moment.Moment;
```
PanopticaRising, line 315: Why is this date a Moment when `timeOfSubmission` is a Date? I feel like we should keep them consistent.

red-tt-bear: yeah i'm back and forth on this, the reason is that solution date is calculated from moment, so to do this I would have to convert from moment back to date, pass it in and the function would convert it to a moment

The other fields come from the database model and all three are used as moment in this function

Instead of changing this to a date I could change timeofSubmission into a moment I suppose

src/features/courses/course-route.ts
red-tt-bear: We cannot, I think there is a comment on the constant but this error can never be triggered, middleware is actually the problem here

The authentication middleware's existence enforces that req.session is not nil, however use strict does not know about checks in other closures

You'll see this elsewhere in the application where we have a closure and we have to have a nil check that can never actually be called

src/features/courses/course-controller.ts
```diff
@@ -1050,6 +1157,491 @@ class CourseController {
...
+                    active: true
+                }
+            });
+            
+            const workbooksAndOverrides: (StudentWorkbook | StudentGradeOverride)[] = [...workbooks, ...gradeOverrides];
+            const sortedWorkbooksAndOverrides = workbooksAndOverrides.sort((first: StudentWorkbook | StudentGradeOverride, second: StudentWorkbook | StudentGradeOverride): number => {
+                const getDate = (object: StudentWorkbook | StudentGradeOverride): Date => {
+                    if (object instanceof StudentWorkbook) {
+                        return object.time;
+                    } else {
```
PanopticaRising, line 1506: This probably would've been cleaner as an elseif and an else.

src/features/courses/course-controller.ts
```diff
@@ -1050,6 +1157,491 @@ class CourseController {
...
+                        if (!(object instanceof StudentGradeOverride)) {
+                            logger.error('Invalid type, should be StudentWorkbook | StudentGradeOverride');
+                        }
+                        // could always be createdAt...
+                        // time for workbook is redundant and should always be the same as created at
+                        // That being said
+                        return object.createdAt;
+                    }
+                };
+                return getDate(first).getTime() - getDate(second).getTime();
```
PanopticaRising, line 1516: Does this take/need to take into account the date? I've seen getTime in a couple of places.

src/utilities/grading-helper.ts
```diff
@@ -0,0 +1,231 @@
...
+export const determineGradingRationale = ({
+    endDate,
+    deadDate,
+    locked,
+
+    overallBestScore,
+    numAttempts,
+
+    maxAttempts,
+
```
PanopticaRising, line 85: Is this spacing intentional?

src/utilities/grading-helper.ts
```diff
@@ -0,0 +1,231 @@
...
+    // This should be the same as doing else, being extra explicity to be careful
+    // Also typescript complains if this is else because `UNKNOWN` is not possible below
+    else if (!isCompleted && !areSolutionsAvailable) {
+        willTrackAttemptReason = WillTrackAttemptReason.YES;
+    } else {
+        logger.error('An error occurred determining whether or not to keep the attempt');
+        willTrackAttemptReason = WillTrackAttemptReason.UNKNOWN;
+    }
+
+    let willGetCreditReason: WillGetCreditReason = WillGetCreditReason.UNKNOWN;
```
PanopticaRising, line 121: I feel like this might be better as a conversion function.

PanopticaRising (approved):

doombot-exe (approved):

src/utilities/grading-helper.ts
red-tt-bear: yes

src/utilities/grading-helper.ts
```diff
@@ -0,0 +1,231 @@
...
+                result.gradeUpdates.bestScore = result.gradeUpdates.overallBestScore;
+                result.gradeUpdates.legalScore = result.gradeUpdates.overallBestScore;
+                result.gradeUpdates.partialCreditBestScore = result.gradeUpdates.overallBestScore;
+                // if it was overwritten to be better use that max value
+                result.gradeUpdates.effectiveScore =
+                    newScore > studentGrade.effectiveScore ?
+                        newScore : undefined;
+            } else if (gradingRationale.willGetCreditReason === WillGetCreditReason.YES_BUT_PARTIAL_CREDIT) {
+                // Partial credit
+                const partialCreditScalar = 0.5;
```
red-tt-bear, line 216: Maybe move to constants

README.md
```diff
@@ -72,4 +72,25 @@
...
+
+##### Logging Levels
+* ERROR
+* WARN
+* INFO
+* HTTP
+* VERBOSE
+* DEBUG
+* SILLY
+* null
```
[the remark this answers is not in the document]
red-tt-bear: logging level of null is don't use the logger, it's documented in the readme, I can see how that'd be confusing, I figured it was a "null" logger

src/utilities/grading-helper.ts
red-tt-bear: What do you mean by conversion function?
I feel like separating out the logic into its own function would be better too if that is what you are saying

PR #150, opened 2020-10-01
Feature/add individual extensions

PanopticaRising:


src/database/models/course-ww-topic-question.ts
```diff
@@ -121,4 +128,5 @@ CourseWWTopicQuestion.init({
 
 import CourseTopicContent from './course-topic-content';
 import StudentGrade from './student-grade';
-import CurriculumWWTopicQuestion from './curriculum-ww-topic-question';
+import CurriculumWWTopicQuestion from './curriculum-ww-topic-question';import StudentTopicQuestionOverride from './student-topic-question-override';
```
red-tt-bear, line 131: This should fail linting, please add new lines here

src/features/courses/course-controller.ts
```diff
@@ -72,11 +74,26 @@ class CourseController {
         });
     }
 
-    getTopicById(id: number): Promise<CourseTopicContent> {
+    getTopicById(id: number, userId: number | undefined = undefined): Promise<CourseTopicContent> {
```
red-tt-bear, line 77: Typically if we have multiple parameters we use an object
Also instead of `userId: number | undefined = undefined` you should just do `userId?: number`

red-tt-bear: Nit ^

src/features/courses/course-controller.ts
```diff
@@ -72,11 +74,26 @@ class CourseController {
...
+        const include = [];
+        if (userId !== undefined) {
```
red-tt-bear, line 79: Nit: I prefer _.isNil, even though right now it can't be null if we ever made so the case in the future it would "just work"
If null ever did get in here it would be a problem right?
Or maybe we do want it this way because there is a reason that null should get through if we allow it in the future?

src/features/courses/course-controller.ts
```diff
@@ -360,6 +407,12 @@ class CourseController {
         });
     }
 
+    async extendTopicForUser(options: ExtendTopicForUserOptions): Promise<UpsertResult<StudentTopicOverride>> {
+        return appSequelize.transaction(async () =>  {
+            return await courseRepository.extendTopicByUser(options);
```
red-tt-bear, line 412: Nit: unnecessary await

src/features/courses/course-controller.ts
```diff
@@ -848,6 +901,16 @@ class CourseController {
         };
     }
 
+    async extendQuestionForUser(options: ExtendTopicQuestionForUserOptions): Promise<UpsertResult<StudentTopicQuestionOverride>> {
+        return appSequelize.transaction(async () =>  {
```
red-tt-bear, line 905: Nit: unecessary async

src/features/courses/course-controller.ts
```diff
@@ -959,8 +1022,46 @@ class CourseController {
                 studentWorkbook: null
             };
         }
-        const question: CourseWWTopicQuestion = await studentGrade.getQuestion();
-        const topic: CourseTopicContent = await question.getTopic();
+        const question: (CourseWWTopicQuestion) = await studentGrade.getQuestion(
```
red-tt-bear, line 1025: Why is the type in parenthesis?

src/features/courses/course-controller.ts
```diff
@@ -959,8 +1022,46 @@ class CourseController {
...
+                attributes: ['userId', 'startDate', 'endDate', 'deadDate'],
+                required: false,
+                where: {
+                    active: true,
+                    userId: options.userId
+                }
+            }]
+        });
+        
+        if ((topic as any)?.studentTopicOverride?.length === 1) {
```
red-tt-bear, line 1054: Not Nit
Linter should have yelled at you for using any
You should add studentTopicOverride to the topic class, then you don't need to cast
Also if you do break linting rules you should add the ignore comment and add a comment as to why you ignored the rule

red-tt-bear: Like you did below 🤣 the below one I can see being a pain and can understand pushing that off because of the `assign`

src/features/courses/course-repository.ts
```diff
@@ -263,6 +263,36 @@ class CourseRepository {
         }
     }
 
+    async extendTopicByUser(options: ExtendTopicForUserOptions): Promise<UpsertResult<StudentTopicOverride>> {
```
red-tt-bear, line 266: NIT: I prefer if, else here
When I first read it I was going to comment it should be in a transaction since two sql commands are run (create and update)
After looking a little harder I noticed that the first if returned
I prefer being explicit because if you ever did want some code to run regardless then you could just add it after the if, else block

red-tt-bear: That's just my preference though ^ no need to change it if you don't want

src/features/courses/course-repository.ts
```diff
@@ -263,6 +263,36 @@ class CourseRepository {
...
+                returning: true,
+            });
+            return {
+                createdNewEntry: false,
+                updatedCount: updates[0],
+                updatedRecords: updates[1],
+            };
+        } catch (e) {
+            this.checkCourseTopicError(e);
+            throw new WrappedError(Constants.ErrorMessage.UNKNOWN_APPLICATION_ERROR_MESSAGE, e);
```
red-tt-bear, line 292: Might be more clear to make this message be something like:
`Could not extend topic`
That message is for us and we could use it at a glance to tell what feature is broken in the logs
I tend to keep `UNKNOWN_APPLICATION_ERROR_MESSAGE` for issues that should never happen, i.e. nulls/undefineds that have already been check but with strict it has to be checked in the callback again

src/features/courses/course-repository.ts
```diff
@@ -483,6 +529,38 @@ class CourseRepository {
             throw new WrappedError('Could not create StudentTopicQuestionOverride', e);
         }
     }
+
+    async extendTopicQuestionByUser(options: ExtendTopicQuestionForUserOptions): Promise<UpsertResult<StudentTopicQuestionOverride>> {
+        try {
+            console.log(options);
```
red-tt-bear, line 535: Remove this console.log

src/features/courses/course-repository.ts
```diff
@@ -483,6 +529,38 @@ class CourseRepository {
...
+            const found = await StudentTopicQuestionOverride.findOne({where: {...options.where, active: true}});
+            if (!found) {
+                const newExtension = await StudentTopicQuestionOverride.create({...options.where, ...options.updates}, {validate: true});
```
red-tt-bear, line 538: Not really pr feedback but a topic for conversation
I used spread but realized that typescript starts treating the type as any, it doesn't actually check if it's a valid spread
I think in the future we should probably move away from using spread operator in typescript

src/features/courses/course-route.ts
```diff
@@ -361,6 +385,27 @@ router.put('/question/grade/:id',
         }));
     }));
 
+router.put('/question/extend',
+authenticationMiddleware,
+validate(extendCourseTopicQuestionValidation),
+asyncHandler(async (req: RederlyExpressRequest<any, ExtendCourseTopicQuestionRequest.body, unknown, any, unknown>, _res: Response, next: NextFunction) => {
+    const query = req.query as ExtendCourseTopicQuestionRequest.query;
+    const body = req.body as ExtendCourseTopicQuestionRequest.body;
+    try {
```
red-tt-bear, line 394: We have this mistake a few different places in the application
We don't need a try catch because `asyncHandler` does that for us

red-tt-bear: Nit ^

src/features/courses/course-route.ts
```diff
@@ -438,15 +483,24 @@ router.get('/question/:id',
...
+            if (req.query.userId) {
+                question = await courseController.getQuestionWithoutRenderer({
+                    id: params.id,
+                    userId: req.query.userId,
+                });
+            } else {
+                // TODO handle not found case
+                question = await courseController.getQuestion({
+                    questionId: params.id,
+                    // TODO: Is there a better way to handle the authentication for this?
```
red-tt-bear, line 496: What do you mean? I think grabbing from the session is probably the best (and most secure way)
You don't have to check as many permissions if it comes from the session (i.e. if we allowed another user to fetch a question as if they were the student like a professor seeing what they see)

src/features/courses/course-route.ts
```diff
@@ -438,15 +483,24 @@ router.get('/question/:id',
...
-            const question = await courseController.getQuestion({
-                questionId: params.id,
-                userId: session.userId,
-                formURL: req.originalUrl,
-                role: user.roleId,
-                readonly: req.query.readonly,
-                workbookId: req.query.workbookId,
-            });
+            let question;
+            if (req.query.userId) {
```
red-tt-bear, line 487: Instead of relying on the userId we should probably add a query flag to fetch from renderer

red-tt-bear (changes requested):

src/features/courses/course-controller.ts
PanopticaRising: I chose not to use an object to minimize impact to other features. Optional parameters made more sense.

Interestingly, I can't initialize it to undefined with the questionmark. `Parameter cannot have question mark and initializer.`
I thought I tried the way you suggested before and it didn't work, but it seems like it is now. Thanks!

src/features/courses/course-controller.ts
PanopticaRising: I think that, in terms of getting a topic with a userId associated that null probably means the same as undefined, so I'll make that change.

src/features/courses/course-controller.ts
PanopticaRising: Thanks, I should've checked how you were handling it over there.

src/features/courses/course-repository.ts
PanopticaRising: Should we have a constants file for error messages?

PanopticaRising: (Besides the seemingly generic ones in Constants.ErrorMessage)

src/features/courses/course-repository.ts
PanopticaRising: I don't think that's a good reason to avoid using it. There might be better support in later versions of Typescript that we could benefit from, but I don't think this is a good enough reason to avoid using it.

src/features/courses/course-route.ts
PanopticaRising: That was actually a slightly out-of-date comment. It was meant to be about using `req.query.userId` in the conditional above this.

src/features/courses/course-controller.ts
```diff
@@ -360,6 +407,12 @@ class CourseController {
         });
     }
 
+    async extendTopicForUser(options: ExtendTopicForUserOptions): Promise<UpsertResult<StudentTopicOverride>> {
+        return appSequelize.transaction(async () =>  {
```
red-tt-bear, line 411: Nit: doesn't need the async now ;)

src/features/courses/course-repository.ts
```diff
@@ -263,6 +263,39 @@ class CourseRepository {
...
+                });
+
+                return {
+                    createdNewEntry: false,
+                    updatedCount: updates[0],
+                    updatedRecords: updates[1],
+                };
+            }
+        } catch (e) {
+            this.checkCourseTopicError(e);
```
red-tt-bear, line 294: Is this the correct check here? Doesn't this method check for topic constraint errors rather than override errors?

src/features/courses/course-repository.ts
red-tt-bear: The point of typescript is to protect you from using incorrect typing, using this you could pass a string as a number etc.
I feel like anything that circumvents typescript is a big problem as you add more and more distrust in the types
i.e. on the front end we don't check types from the api calls so dates come as string and caused bugs (the very thing typescript tries to avoid)

Not a now issue but we should talk about it further

src/features/courses/course-route.ts
```diff
@@ -214,6 +202,18 @@ router.get('/questions',
             userId = userIdInput;
         }
 
+        let topic: CourseTopicContent | null = null;
+        if(!_.isNil(req.query.courseTopicContentId)) {
+            topic = await courseController.getTopicById(req.query.courseTopicContentId, userId);
+            if (new Date().getTime() < topic.startDate.getTime()) {
```
red-tt-bear, line 208: Nit: We should keep this consistent, if we use moment in some places we should stick with moment

src/features/courses/course-route.ts
```diff
@@ -225,6 +225,30 @@ router.get('/questions',
         }));
     }));
 
+router.put('/topic/extend',
+    authenticationMiddleware,
+    validate(extendCourseTopicForUserValidation),
+    // This is due to a typescript issue where the type mismatches extractMap
```
red-tt-bear, line 231: Nit: Copy paste comment, this isn't relevant

src/features/courses/course-route.ts
```diff
@@ -225,6 +225,30 @@ router.get('/questions',
...
+    asyncHandler(
+        async (req: RederlyExpressRequest<ExtendCourseTopicForUserRequest.params, ExtendCourseTopicForUserRequest.body, ExtendCourseTopicForUserRequest.query, unknown>, _res: Response, next: NextFunction) => {
+        const query = req.query as ExtendCourseTopicForUserRequest.query;
+        const body = req.body as ExtendCourseTopicForUserRequest.body;
+        try {
```
red-tt-bear, line 236: Nit: don't technically need try catch

src/features/courses/course-route.ts
red-tt-bear: Nit ^

src/features/courses/course-route.ts
```diff
@@ -361,6 +385,25 @@ router.put('/question/grade/:id',
         }));
     }));
 
+router.put('/question/extend',
+authenticationMiddleware,
+validate(extendCourseTopicQuestionValidation),
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
red-tt-bear, line 391: Missing the comment that was copy pasted above

src/features/courses/course-route.ts
```diff
@@ -575,6 +626,21 @@ router.get('/',
         }
     }));
 
+// This returns information about a specific topic. Currently, it only 
+// returns extension information if a specific user is passed.
+router.get('/topic/:id', 
+    authenticationMiddleware,
+    validate(getTopicValidation),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
red-tt-bear, line 634: Missing explanation comment

src/features/courses/course-controller.ts
```diff
@@ -959,8 +1022,46 @@ class CourseController {
...
+                        where: {
+                            active: true,
+                            userId: options.userId
+                        }
+                }]
+            }
+        );
+        
+        
+        const topic: CourseTopicContent = await question.getTopic(            {
```
red-tt-bear, line 1041: Floating bracket?

src/features/courses/course-route.ts
red-tt-bear: Talked to @PanopticaRising not doing this now

src/features/courses/course-controller.ts
red-tt-bear: yeah you can't do:
```
userId?: number = undefined
```
it has to be
```
userId?: number
```

red-tt-bear: yeah for now I think it's fine, however we should try to stay consistent in the future, it's really my fault for not making it an object in the first place

src/features/courses/course-repository.ts
PanopticaRising: You're right, I missed that when I copied and modified the original function.

src/features/courses/course-route.ts
PanopticaRising: This was old code, but I agree. I'll patch it up.

src/features/courses/course-repository.ts
red-tt-bear: We could if it grew larger but I don't think we need it
In this case, it's an internal server error so this message is just for us
Additionally, this error message shouldn't be reused

src/features/courses/course-repository.ts
```diff
@@ -483,6 +531,37 @@ class CourseRepository {
...
+                },
+                returning: true,
+            });
+            return {
+                createdNewEntry: false,
+                updatedCount: updates[0],
+                updatedRecords: updates[1],
+            };
+        } catch (e) {
+            this.checkCourseTopicError(e);
```
red-tt-bear, line 560: Is this another missed copy paste?

red-tt-bear (approved):

PR #157, opened 2020-10-14
Feature/red 116 exam settings

PanopticaRising:


src/routes.ts
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 import express = require('express');
 const router = express.Router();
-import { version } from '../package.json';
+// import { version } from '../package.json';
```
red-tt-bear, line 5: @PanopticaRising 😞 
Please uncomment temporary fixes after they are merged in...

red-tt-bear:
Package-lock.json has changes but package.json does? That doesn't seem right

red-tt-bear (changes requested):
Pointed out some obvious stuff, this cannot be reviewed the way that it is with a massive amount of changes between two devs

PR #158, opened 2020-10-14
Feature/red 123 exam versioning

doombot-exe:
draft PR - untested
includes:
* latest updates to assessment table models + migration
* create assessment version
* handle assessment submission + grading + workbooks


red-tt-bear (changes requested):
Pointed out some obvious stuff, this cannot be reviewed the way that it is

src/database/models/student-grade.ts
```diff
@@ -1,257 +1,278 @@
...
-    // // Here we associate which actually populates out pre-declared `association` static and other methods.
-    // User.hasMany(Session, {
-    //   sourceKey: 'id',
-    //   foreignKey: 'user_id',
-    //   as: 'user' // this determines the name in `associations`!
-    // });
+    static getVersion = (obj: StudentGradeInterface, version: StudentGradeInstanceGradeOverridesInterface): StudentGradeInterface => {
+        // Avoid cyclic dependencies
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        return _.assign({}, obj, version); // will override randomSeed, numAttempts, and currentProblemState ** do we want scores too?
```
red-tt-bear, line 73: This isn't doing a pick thus it should be override fields like id as well? I would check because this could subtly make some really major bugs

red-tt-bear: Granted it shouldn't be so bad because it returns the stripped model

src/database/models/topic-assessment-info.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+export default class TopicAssessmentInfo extends Model implements TopicAssessmentInfoInterface {
+    public id!: number;
+    public courseTopicContentId!: number;
+    public duration!: number; // enforce IN MINUTES
+    public hardCutoff!: boolean;
+    public maxGradedAttemptsPerVersion!: number;
+    public maxVersions!: number;
+    public versionDelay!: number; // for consistency do we also force MINUTES here?
```
red-tt-bear, line 26: I think so, should we append `minutes` to the field name to make it clear?

src/features/courses/course-controller.ts
```diff
@@ -2347,7 +2382,26 @@ class CourseController {
                     ['problemNumber', 'ASC'],
                 ]
             };
-            return await CourseWWTopicQuestion.findAll(findOptions);
+            const questions = await CourseWWTopicQuestion.findAll(findOptions);
+            if (!_.isNil(courseTopicContentId)){
+                const topic = await this.getTopicById({id: courseTopicContentId});
+                // TODO remove assessment hardcoding
```
red-tt-bear, line 2388: Yup we should change this to enum

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2634,243 @@ class CourseController {
             throw new WrappedError('Could not create new student grade', e);
         }
     }
+
+    // does this actually belong in course-repository?
```
red-tt-bear, line 2638: typically you'll have a controller method and a repository method, the content of this would go in the repository and a method just calling the repository would be here

The reason for this is to keep a clean responsibility chain
Routes ==> Controller ==> Repository

Also it gives a centralized point if we wanted to do application logic on the create (i.e. check for fields maybe)

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2634,243 @@ class CourseController {
...
+        const { topicId, userId } = options;
+        const topic = await courseRepository.getCourseTopic({id: topicId});
+        const topicInfo = await topic.getTopicAssessmentInfo(); // order by startDate 
+
+        const questions = await courseRepository.getQuestionsFromTopicId({id:topicId});
+        const startTime = moment().add(1,'minute'); // 1 minute should cover any delay in creating records before a student can actually begin
+
+        let endTime = moment(startTime).add(topicInfo.duration, 'minutes');
+        if (topicInfo.hardCutoff) {
+            endTime = moment.min(moment(topic.deadDate), moment(startTime).add(topicInfo.duration, 'minutes'));
```
red-tt-bear, line 2670: We have a convenient extension toMoment
This could be:
endTime = moment.min(topic.deadDate.toMoment(), startTime.toMoment().add(topicInfo.duration, 'minutes'));

I think this looks cleaner
The main reason for the extension is for null chaining too

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2634,243 @@ class CourseController {
...
+
+        results.forEach((result: SubmittedAssessmentResultContext) => {
+            const { questionResponse, grade, instance, weight } = result;
+            totalScore += weight * questionResponse.problem_result.score;
+            bestVersionScore += weight * instance.scoreForBestVersion;
+            bestOverallVersion += weight * grade.bestScore;
+            problemScores[result.instance.problemNumber] = result.questionResponse.problem_result.score * result.weight;
+        });
+
+        problemScores[0] = totalScore; // scores indexed by problem number, total at [0]
```
red-tt-bear, line 2765: I don't think problem scores should be an array, there is no guarantee that problem numbers have no gaps and are 1 based

This should never be the case but then again we saw issues with this assumption when we first did create course

Instead I would rather problemScores was of type `{[key: number]: number}` (dictionary of number to number)

The key for total score should be total score for readability

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2634,243 @@ class CourseController {
...
+        return {problemScores, bestVersionScore: bestVersionScore, bestOverallVersion};
+    }
+
+    async submitAssessmentAnswers(studentTopicAssessmentInfoId: number, wasAutoSubmitted: boolean): Promise<SubmitAssessmentAnswerResult> {
+        const studentTopicAssessmentInfo = await this.getStudentTopicAssessmentInfoById(studentTopicAssessmentInfoId);
+        if (studentTopicAssessmentInfo.numAttempts >= studentTopicAssessmentInfo.maxAttempts) {
+            throw new WrappedError('Cannot submit assessment answers when there are no attempts remaining'); // sanity check, shouldn't happen
```
red-tt-bear, line 2772: How should this be returned to the user, the way it is now will be an internal server error, however it seems like it should be a bad request

src/features/courses/course-route.ts
```diff
@@ -205,17 +201,43 @@ router.get('/questions',
...
+                    next(Boom.badRequest('Topic is an assessment, but does not have corresponding assessment info. This should never happen.'));
+                    return;
+                }
+
+                // get version info - descending by startTime
+                const versions = await courseController.getStudentTopicAssessmentInfo({userId: user.id, topicId: req.query.courseTopicContentId});
+
+                if (
+                    topic.topicAssessmentInfo.hideProblemsAfterFinish &&
+                    !_.isNil(versions[0]) && (
```
red-tt-bear, line 230: We need an array check

src/features/courses/course-route.ts
```diff
@@ -275,6 +297,72 @@ router.put('/topic/:id',
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, CreateAssessmentVersionRequest.body, CreateAssessmentVersionRequest.query>, _res: Response, next: NextFunction) => {
+        const params = req.params as CreateAssessmentVersionRequest.params;
+        if (_.isNil(req.session)) {
+            throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+        const user = await req.session.getUser();
+
+        let topic = await courseController.getTopicById({id: params.id}); // includes TopicOverrides
+        const topicInfo = await courseController.getTopicAssessmentInfoByTopicId({ topicId: topic.id, userId: user.id }); // ordered by startDate, includes overrides
```
red-tt-bear, line 313: I know @PanopticaRising like keeping these without spacing, however I personally do not like this

The pro is brevity keeping the code looking shorted

The con IMO is that it is less readable (easier to read a list with newlines) and more likely to have merge conflicts

src/features/courses/course-route.ts
```diff
@@ -275,6 +297,72 @@ router.put('/topic/:id',
...
+            }
+        }
+
+        // finally - *no one* should be able to create a version if there is already one "open"
+        // TODO: add check that the "open" version hasn't been flagged as "isClosed"
+        if (studentVersions[0] && new Date().getTime() < studentVersions[0].endTime.getTime()) {
+            next(Boom.badRequest('You already have an open version of this assessment.'));
+        }
+
+        try {
```
red-tt-bear, line 353: Don't need try catch

red-tt-bear (changes requested):

src/features/courses/course-controller.ts
red-tt-bear: {[key: string]: number}
{[key: string | number]: number}
{[key: any]: number} // Don't do this!

src/database/models/student-grade.ts
doombot-exe: StudentGradeInstanceGradeOverridesInterface only includes `randomSeed` and `currentProblemState`

doombot-exe: I imitated what you did for overrides elsewhere

src/database/models/student-grade-instance.ts
```diff
@@ -0,0 +1,232 @@
...
+    // numAttempts: number;
+    // This is a jsonb field so it could be any (from db)
+    // Submitted in workbook used any so I'm going to keep it consistent here
+    // If this is used for form data we will never know any info about what keys are available
+    // Might make sense to make this an unknown type since I don't think we will ever access the types
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    currentProblemState: any;
+}
+
+export interface StudentGradeInstanceInterface {
```
red-tt-bear, line 21: This should extend the above two interfaces ^

red-tt-bear: It would reduce duplicate code as well for allowing polymorphism

src/database/models/student-grade.ts
red-tt-bear: Gotcha, it blended in, looks good

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2641,250 @@ class CourseController {
...
+        let bestVersionScore = 0;
+        let bestOverallVersion = 0;
+        let problemScores: { [key: string]: number } = {};
+
+        results.forEach((result: SubmittedAssessmentResultContext) => {
+            const { questionResponse, grade, instance, weight } = result;
+            totalScore += weight * questionResponse.problem_result.score;
+            bestVersionScore += weight * instance.scoreForBestVersion;
+            bestOverallVersion += weight * grade.bestScore;
+            problemScores = {[result.instance.problemNumber.toString(10)]: result.questionResponse.problem_result.score * result.weight};
```
red-tt-bear, line 2776: This isn't doing what your thinking your doing (I think new lines would actually help make it clearer).
You're setting `problemScores` equal to a new object with a single fields set, thus none of the previous scores were kept, and you override it again below so only the total score would be kept

What you want to do:
```typescript
problemScores[result.instance.problemNumber.toString(10)] = result.questionResponse.problem_result.score * result.weight
```

Notice that the braces are gone and the key is used in brackets on the reference side

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2641,250 @@ class CourseController {
...
+        });
+
+        problemScores = {total: totalScore}; // scores indexed by problem number, total at [0]
```
red-tt-bear, line 2779: The comment here is no longer appropriate

src/features/courses/course-route.ts
```diff
@@ -275,6 +303,72 @@ router.put('/topic/:id',
         }
     }));
 
+router.get('/assessment/topic/:id/start',
```
red-tt-bear, line 306: This route has too much business logic

red-tt-bear (changes requested):
The submit submission call does not return the itemized results

src/database/models/student-grade-instance.ts
doombot-exe: cool - I didn't realize that was possible

src/features/courses/course-controller.ts
doombot-exe: I was suspicious of this, but couldn't work out how to use a variable as a key name...

src/features/courses/course-route.ts
doombot-exe: handled - moved all the checks to a function that returns boolean + message

src/features/courses/course-controller.ts
```diff
@@ -2347,7 +2383,26 @@ class CourseController {
                     ['problemNumber', 'ASC'],
                 ]
             };
-            return await CourseWWTopicQuestion.findAll(findOptions);
+            const questions = await CourseWWTopicQuestion.findAll(findOptions);
+            if (!_.isNil(courseTopicContentId)){
+                const topic = await this.getTopicById({id: courseTopicContentId});
+                // TODO remove assessment hardcoding
+                if (topic.topicTypeId === 2 && !_.isNil(userId)) {
+                    questions.asyncForEach(async (question) => {
```
red-tt-bear, line 2391: You need to await this

red-tt-bear (approved):

src/database/migrations/20200925034242-assessments-setup.ts
```diff
@@ -0,0 +1,457 @@
...
+                    field: 'created_at',
+                    type: DataTypes.DATE,
+                    allowNull: false,
+                },
+            });
+
+            /**
+             * topic_assessment_info
+             */
+            await queryInterface.createTable('topic_assessment_info', {
```
PanopticaRising, line 62: Is there a reason we dropped the `course_` prefix from this but not from the `course_question_assessment_info`?

src/database/models/student-grade-instance.ts
```diff
@@ -0,0 +1,210 @@
...
+    // Don't need references to workbooks since there should be a relatively 1-1 relationship
+    active: boolean;
+    bestIndividualAttemptId: number;
+    bestVersionAttemptId: number;
+}
+
+export default class StudentGradeInstance extends Model implements StudentGradeInstanceInterface {
+    public id!: number;
+    public studentGradeId!: number;
+    // public userId!: number;
```
PanopticaRising, line 45: why is this commented out? Do we need it for the association?

red-tt-bear: Good catch
This part is just decorative for typescript, it gives no functionality.
It would only be a problem if we wanted to access the userId

That being said though your right, why are we commenting this out

src/database/models/student-topic-assessment-override.ts
```diff
@@ -0,0 +1,110 @@
...
+    versionDelay: number | null;
+    active: boolean;
+}
+
+export default class StudentTopicAssessmentOverride extends Model implements StudentTopicAssessmentOverrideInterface {
+    public id!: number;
+    // public courseTopicContentId!: number;
+    public topicAssessmentInfoId!: number;
+    public userId!: number;
+    public duration!: number | null;
```
PanopticaRising, line 20: @red-tt-bear should these use `?` instead of `!`?

red-tt-bear: Nah, this is from sequelize [documentation](https://sequelize.org/master/manual/typescript.html)

red-tt-bear: Sorry I missed what you were saying, yeah I think the duration should have a question mark

src/database/models/topic-assessment-info.ts
```diff
@@ -0,0 +1,161 @@
...
+            as: 'studentTopicAssessmentOverride'
+        });
+
+        TopicAssessmentInfo.hasMany(StudentTopicAssessmentInfo, {
+            foreignKey: 'topicAssessmentInfoId',
+            sourceKey: 'id',
+            as: 'studentTopicAssessmentInfo'
+        });
+
+        // CourseTopicContent.hasMany(CourseWWTopicQuestion, {
```
PanopticaRising, line 67: This should be deleted

src/database/models/student-topic-assessment-override.ts
red-tt-bear: Actually, according to that documentation, this is correct
> public preferredName!: string | null; // for nullable fields

red-tt-bear: I think we should leave it for now and deal with it later as it matches the documentation, I feel like there is a non 0 chance the documentation is wrong

src/features/courses/course-repository.ts
```diff
@@ -414,6 +451,19 @@ class CourseRepository {
...
 
+    async getQuestionsFromTopicId(options: GetQuestionRepositoryOptions): Promise<CourseWWTopicQuestion[]> {
+        const result = await CourseWWTopicQuestion.findAll({
+            where: {
+                courseTopicContentId: options.id,
+                active: true
+            },
+        });
+        if (_.isNil(result)) {
+            throw new NotFoundError('Questions requested from a non-existent topic');
```
PanopticaRising, line 462: Technically, this could mean that there are just no questions, but I don't think it merits changing.

src/features/courses/course-route.ts
```diff
@@ -205,17 +202,51 @@ router.get('/questions',
...
+            if (topic?.topicTypeId === 2 && !_.isNil(userId)) {
+                if (_.isNil(topic.topicAssessmentInfo)){
+                    next(Boom.badRequest('Topic is an assessment, but does not have corresponding assessment info. This should never happen.'));
+                    return;
+                }
+
+                // get version info - descending by startTime unless specific id is included in query
+                let versions: StudentTopicAssessmentInfo[] | undefined;
+                if (_.isNil(req.query.studentTopicAssessmentInfoId)) {
+                    versions = _.orderBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'],'desc');
```
PanopticaRising, line 229: If the other one wasn't working, this one might also need a fix
```suggestion
                    versions = _.orderBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'], ['desc']);
```

src/features/courses/course-route.ts
```diff
@@ -205,17 +202,51 @@ router.get('/questions',
...
+                }
+
+                if (
+                    topic.topicAssessmentInfo.hideProblemsAfterFinish && (
+                        moment().isAfter(moment(versions[0].endTime)) ||
+                        topic.topicAssessmentInfo.maxGradedAttemptsPerVersion <= versions[0].numAttempts // TODO: replace with versions[0].isFinished
+                    ) &&
+                    user.roleId === Role.STUDENT
+                ) {
+                    next(httpResponse.Ok('You have finished this version of the assessment and you are blocked from seeing the problems.', {questions: null, topic}));
```
PanopticaRising, line 246: Better verbiage might be "Problems on this exam many not be viewed until the exam has finished for all students ${endDate}"

src/features/courses/course-types.ts
```diff
@@ -25,10 +27,11 @@ export interface CourseListOptions {
     };
 }
 
-export interface GetQuestionOptions {
-    userId: number;
-    questionId: number;
-}
+// this is defined twice
```
PanopticaRising, line 30: Remove unused code

src/features/courses/course-controller.ts
```diff
@@ -2522,6 +2654,335 @@ class CourseController {
...
+            topic = topic.getWithOverrides(topic.studentTopicOverride[0]) as CourseTopicContent;
+        }
+
+        // restrictions on start time, maximum versions, and version delay only apply to students
+        if (user.roleId === Role.STUDENT) {
+            // if the assessment has not yet started, students cannot generate a version
+            if (new Date().getTime() < topic.startDate.getTime()) {
+                message = `The topic "${topic.name}" has not started yet.`;
+                userCanStartNewVersion = false;
+            } else
```
PanopticaRising, line 2834: I think we should inline `else if`s together. Personally, it makes it harder for me to see them as a contiguous block- I initially read these as noops.

PanopticaRising (approved):

src/features/courses/course-controller.ts
red-tt-bear: 💯 (some reactions aren't available to attach)

src/features/courses/course-route.ts
red-tt-bear: Probably true, although I don't know if we use this message from the server, most of the time we don't
That being said it is still good to keep it clean

src/features/courses/course-repository.ts
red-tt-bear: Would no questions result in nil or empty array?

src/database/models/student-grade-instance.ts
doombot-exe: grade instance -> student grade -> user or grade instance -> student topic assessment info -> user

there are several fields commented out here, not just userId, we should dump them all

src/database/models/topic-assessment-info.ts
doombot-exe: ehhh, I think that's important for UI -- but what if we change to nanoseconds in the future?

src/features/courses/course-repository.ts
doombot-exe: you're right - I think this check is not enough, length === 0 instead? (And a more accurate error message)

src/features/courses/course-route.ts
doombot-exe: I want to be able to use these messages in the modal that is displayed to students on the front end.

src/features/courses/course-controller.ts
doombot-exe: [Noops](https://i.pinimg.com/originals/0b/51/6e/0b516e1113ca5f8ebb094af23e47a5ce.jpg)

src/features/courses/course-route.ts
doombot-exe: already fixed in the next branch

PR #159, opened 2020-10-15
Feature/red 116 exam settings

PanopticaRising:


src/database/models/topic-assessment-info.ts
```diff
@@ -94,7 +94,10 @@ TopicAssessmentInfo.init({
         field: 'topic_assessment_info_max_graded_attempts_per_version',
         type: DataTypes.INTEGER,
         allowNull: false,
-        defaultValue: 0
+        defaultValue: 1,
+        validate: {
```
red-tt-bear, line 98: @doombot-exe on practice exams can't this be 0 or -1?

src/features/courses/course-controller.ts
```diff
@@ -81,7 +81,7 @@ class CourseController {
         });
     }
 
-    getTopicById(id: number, userId?: number): Promise<CourseTopicContent> {
+    getTopicById({id, userId, includeQuestions}: {id: number; userId?: number; includeQuestions?: boolean}): Promise<CourseTopicContent> {
```
red-tt-bear, line 84: Nit: inlining types is not great, we can move that later

src/features/courses/course-repository.ts
```diff
@@ -209,6 +209,16 @@ class CourseRepository {
                 id: options.id,
                 active: true
             },
+            include: [
```
red-tt-bear, line 212: Should this include be a passed in option so it is not always executed?

PanopticaRising: I don't think so. Would we add a new flag for 'includeExams'? I don't really see what the benefit to that would be.

red-tt-bear: less queries if you were fetching the topic for some other reason

red-tt-bear (approved):

src/database/models/topic-assessment-info.ts
doombot-exe: yes, but the default should be 1 attempt - as that is likely the most commonly preferred setting

src/features/courses/course-repository.ts
doombot-exe: we'd have to otherwise check that a topic is an exam anyhow - I certainly make use of this include being here when I request a topic.

doombot-exe (approved):

PR #161, opened 2020-10-20
RED-144 updates for submit exam + response

doombot-exe:
preliminary pull request to compliment branch of the same name on the frontend:
- students will be asked to take a new version if they don't have one open
- students problem states save and can be submitted
- grade results are shown in a modal according to the exam settings
- after an exam is over, problems will be hidden and a new version will need to be started
- backend will not comply when the delay hasn't passed yet, but the front end isn't ready for that
- exceeding the number of allowed graded attempts per version makes the back end err, but the message doesn't make it beyond STDERR
- there are only backend checks against exceeding the allowed number of versions
- an exam version cannot be ended early yet

### DIFF
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -16,7 +16,7 @@ import logger from '../../utilities/logger';
 import sequelize = require('sequelize');
 import WrappedError from '../../exceptions/wrapped-error';
 import AlreadyExistsError from '../../exceptions/already-exists-error';
-import { GetTopicsOptions, CourseListOptions, UpdateUnitOptions, UpdateTopicOptions, EnrollByCodeOptions, GetGradesOptions, GetStatisticsOnQuestionsOptions, GetStatisticsOnTopicsOptions, GetStatisticsOnUnitsOptions, GetQuestionOptions, GetQuestionResult, SubmitAnswerOptions, SubmitAnswerResult, FindMissingGradesResult, GetQuestionsOptions, GetQuestionsThatRequireGradesForUserOptions, GetUsersThatRequireGradeForQuestionOptions, CreateGradesForUserEnrollmentOptions, CreateGradesForQuestionOptions, CreateNewStudentGradeOptions, UpdateQuestionOptions, UpdateCourseOptions, MakeProblemNumberAvailableOptions, MakeUnitContentOrderAvailableOptions, MakeTopicContentOrderAvailableOptions, CreateCourseOptions, CreateQuestionsForTopicFromDefFileContentOptions, DeleteQuestionsOptions, DeleteTopicsOptions, DeleteUnitsOptions, GetCalculatedRendererParamsOptions, GetCalculatedRendererParamsResponse, UpdateGradeOptions, DeleteUserEnrollmentOptions, ExtendTopicForUserOptions, GetQuestionRepositoryOptions, ExtendTopicQuestionForUserOptions, GradeOptions, ReGradeStudentGradeOptions, ReGradeQuestionOptions, ReGradeTopicOptions, SetGradeFromSubmissionOptions, CreateGradeInstancesForAssessmentOptions, CreateNewStudentGradeInstanceOptions, CreateNewStudentTopicAssessmentInfoOptions, GetStudentTopicAssessmentInfoOptions, GetTopicAssessmentInfoByTopicIdOptions, SubmittedAssessmentResultContext, SubmitAssessmentAnswerResult, ScoreAssessmentResult, UserCanStartNewVersionOptions, UserCanStartNewVersionResult, UpdateGradeInstanceOptions } from './course-types';
+import { GetTopicsOptions, CourseListOptions, UpdateUnitOptions, UpdateTopicOptions, EnrollByCodeOptions, GetGradesOptions, GetStatisticsOnQuestionsOptions, GetStatisticsOnTopicsOptions, GetStatisticsOnUnitsOptions, GetQuestionOptions, GetQuestionResult, SubmitAnswerOptions, SubmitAnswerResult, FindMissingGradesResult, GetQuestionsOptions, GetQuestionsThatRequireGradesForUserOptions, GetUsersThatRequireGradeForQuestionOptions, CreateGradesForUserEnrollmentOptions, CreateGradesForQuestionOptions, CreateNewStudentGradeOptions, UpdateQuestionOptions, UpdateCourseOptions, MakeProblemNumberAvailableOptions, MakeUnitContentOrderAvailableOptions, MakeTopicContentOrderAvailableOptions, CreateCourseOptions, CreateQuestionsForTopicFromDefFileContentOptions, DeleteQuestionsOptions, DeleteTopicsOptions, DeleteUnitsOptions, GetCalculatedRendererParamsOptions, GetCalculatedRendererParamsResponse, UpdateGradeOptions, DeleteUserEnrollmentOptions, ExtendTopicForUserOptions, GetQuestionRepositoryOptions, ExtendTopicQuestionForUserOptions, GradeOptions, ReGradeStudentGradeOptions, ReGradeQuestionOptions, ReGradeTopicOptions, SetGradeFromSubmissionOptions, CreateGradeInstancesForAssessmentOptions, CreateNewStudentGradeInstanceOptions, GetStudentTopicAssessmentInfoOptions, GetTopicAssessmentInfoByTopicIdOptions, SubmittedAssessmentResultContext, SubmitAssessmentAnswerResult, ScoreAssessmentResult, UserCanStartNewVersionOptions, UserCanStartNewVersionResult, UpdateGradeInstanceOptions } from './course-types';
 import { Constants } from '../../constants';
 import courseRepository from './course-repository';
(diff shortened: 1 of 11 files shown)


src/features/courses/course-controller.ts
```diff
@@ -1150,6 +1165,7 @@ class CourseController {
                 formData = gradeInstance.currentProblemState;
                 sourceFilePath = gradeInstance.webworkQuestionPath;
                 problemSeed = gradeInstance.randomSeed;
+                calculatedRendererParameters.outputformat = OutputFormat.ASSESS;
```
red-tt-bear, line 1168: Why not add this to the getCalculatedRendererParams?
Maybe add an option for isAssessment that returns ASSESS

Also this look like too generic of a way of putting this here, if we ever have a non null workbook (like viewing previous submissions, I'm guessing that this is in one of the get grades and I would assume this would break the workbook veiwer (but I could be wrong)

Either way moving the logic into the calculate renderer params should make it more clear

src/features/courses/course-controller.ts
```diff
@@ -2357,19 +2363,33 @@ class CourseController {
...
 
         try {
             const include: sequelize.IncludeOptions[] = [];
+            const subInclude: sequelize.IncludeOptions[] = [];
+
+            if (!_.isNil(studentTopicAssessmentInfoId)) {
+                subInclude.push({
+                    model: StudentGradeInstance,
+                    as: 'gradeInstances',
+                    required: true,
```
red-tt-bear, line 2378: Why is this required true?
If we missing records does it make sense to return them and then log errors with null checks

src/features/courses/course-controller.ts
```diff
@@ -2399,16 +2419,18 @@ class CourseController {
             const questions = await CourseWWTopicQuestion.findAll(findOptions);
             if (!_.isNil(courseTopicContentId)){
                 const topic = await this.getTopicById({id: courseTopicContentId});
-                // TODO remove assessment hardcoding
+                // TODO remove assessment hardcoding -- userId nil-check is for TS
```
red-tt-bear, line 2422: > nil-check is for TS?
What do you mean by this? In other cases it is because of how TS manages scoping, if you nil check and then put it in a closure TS will be wrong and think it is nillable, but in this case it looks like legit concern

src/features/courses/course-controller.ts
```diff
@@ -2399,16 +2419,18 @@ class CourseController {
...
                 if (topic.topicTypeId === 2 && !_.isNil(userId)) {
                     await questions.asyncForEach(async (question) => {
                         if (_.isNil(question.grades)) throw new RederlyExtendedError('Impossible! Found an assessment question without a grade.');
                         const version = await courseRepository.getCurrentInstanceForQuestion({questionId: question.id, userId});
                         question.webworkQuestionPath = version?.webworkQuestionPath ?? question.webworkQuestionPath;
                         question.grades.randomSeed = version?.randomSeed ?? question.grades.randomSeed;
+                        question.problemNumber = version?.problemNumber ?? question.problemNumber;
```
red-tt-bear, line 2429: Should we do this through the assign with the utility function like we have been doing elsewhere?

red-tt-bear (changes requested):
Please move the Assess renderer param to the calculate
Please move required true to false and log nils unless it is intentional

src/features/courses/course-route.ts
```diff
@@ -451,6 +453,32 @@ router.put('/question/grade/:id',
         }));
     }));
 
+router.put('/question/grade/instance/:id',
```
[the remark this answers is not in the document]
doombot-exe: current problem state

src/features/courses/course-controller.ts
```diff
@@ -2730,7 +2753,7 @@ class CourseController {
                 webworkQuestionPath,
                 randomSeed,
                 userId,
-                problemNumber: problemOrder[index],
+                problemNumber: problemOrder[index]+1, // problemOrder starts from 0
```
PanopticaRising, line 2756: We should consider starting to use explicit functions for when we need to increment the problemNumber, so it's clear when we're using the `raw problemNumber` and a `user-facing problemNumber`

src/features/courses/course-route.ts
```diff
@@ -224,22 +225,22 @@ router.get('/questions',
                 }
 
                 // get version info - descending by startTime unless specific id is included in query
-                let versions: StudentTopicAssessmentInfo[] | undefined;
                 if (_.isNil(req.query.studentTopicAssessmentInfoId)) {
-                    versions = _.orderBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'],'desc');
+                    // _.orderBy puts in ascending order
+                    const versions = _.orderBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime']).reverse();
```
PanopticaRising, line 230: https://lodash.com/docs/4.17.15#orderBy `['
```suggestion
                    const versions = _.orderBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'], ['desc']);
```

src/features/courses/course-controller.ts
doombot-exe: if the request includes a specific student_topic_assessment_info_id, then there *must* be associated student grade instances. That ID identifies a specific exam version.

src/features/courses/course-route.ts
doombot-exe: yeah, I think I was confusing _.sortBy and _.orderBy

src/features/courses/course-controller.ts
doombot-exe: Yes, this is why I used problemOrder to stash the (potentially randomized) sequence - I just didn't see the point of mapping x => x+1 instead of doing it here.

src/features/courses/course-controller.ts
doombot-exe: this should be moved - agreed - but adding reliance on topicType to the existing logic was a complication where the concern was breaking existing functionality... will handle this in cleanup.

src/features/courses/course-controller.ts
doombot-exe: yeah, I shouldn't have to check it again, but TS was complaining...

src/features/courses/course-controller.ts
doombot-exe: agreed, but this question object is a sequelize with includes -- so it's not overwriting one of our "fundamental" objects.

red-tt-bear: Yeah, but that is why we do the multiple interface approach and do the assign using _.pick

src/features/courses/course-controller.ts
red-tt-bear: So you are saying that:
since student topic assessment exists then grade instances must exist

I can agree with that, however, we probably need the nil check anyway and if the generation has a mistake then we will catch it in those nil checks

With required true it would return that student the student grade instance didn't exist and then we would have a red herring of a bug

src/features/courses/course-controller.ts
red-tt-bear: Yes you should, this is not in the closure for the userId check
if userId is nil the above if block will not run, since it is no longer in that if statement it could once again be nil

This is a valid TS complaint

src/features/courses/course-controller.ts
red-tt-bear: I'm fine with that

However my concern is more so with adding extra functionality that will change the existing

We've been kicking a lot of stuff downstream and the result is that either:
1. We will forget about it
2. Higher priority tasks will build up between where we will build this into a deeper level of code that is harder to touch
3. Bugs / problems will come of it before we get there (i.e. if we do this outside of the calculate and need it in multiple places but do it differently)

src/features/courses/course-controller.ts
doombot-exe: I am saying that the student-topic-assessment-info and its corresponding student-grade-instance children are created in the same transaction. You cannot have either independently.

So, the recommendation is a nil check on the result of this query before returning?

src/features/courses/course-controller.ts
doombot-exe: Oh right - those are already implemented, I just forgot to take advantage!

src/features/courses/course-controller.ts
red-tt-bear: not before returning, the type has it as an optional field so when you go to access it you would have to nil check there anyway, I'm saying we have one of our classic impossible errors be thrown at that point

Although I agree that currently, it is not possible with the transaction all it takes is either:
* for someone to delete one of the awaits by accident and have it as an orphaned promise thus not respecting the transaction (starting a new one)
* for some change to come in that someone creates these objects independently and do it incorrect

red-tt-bear (approved):

src/features/courses/course-route.ts
```diff
@@ -251,7 +254,8 @@ router.get('/questions',
 
         const questions = await courseController.getQuestions({
             userId: userId,
-            courseTopicContentId: req.query.courseTopicContentId
+            courseTopicContentId: req.query.courseTopicContentId,
+            studentTopicAssessmentInfoId: req.query.studentTopicAssessmentInfoId ?? version?.id,
```
PanopticaRising, line 258: This will cause an error to get thrown if `version` is null, right? So this should be fine?

PR #162, opened 2020-10-20
Feature/red 119 red 120 problem exam settings

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -115,7 +116,15 @@ class CourseController {
                 required: false,
                 where: {
                     active: true,
-                }
+                },
+                include: [{
```
red-tt-bear, line 120: I'm still not a fan of putting more includes in the default getters we already have

It shouldn't hurt anything but for instance if it isn't an assessment we are doing an extra join we don't need

red-tt-bear: No need to change anything for now though

src/features/courses/course-repository.ts
```diff
@@ -553,6 +554,24 @@ class CourseRepository {
                 },
                 returning: true,
             });
+
+            if (updates[0] > 1) {
+                console.error('A single question was expected to update, but multiple update objects are returned.');
```
red-tt-bear, line 559: logger.error, if you do it this way it won't follow logging rules

red-tt-bear (approved):

PR #163, opened 2020-10-21
Feature/red 122 student assessment overrides

red-tt-bear:


src/features/courses/course-route-validation.ts
```diff
@@ -102,13 +102,22 @@ export const updateCourseTopicValidation = {
 export const extendCourseTopicForUserValidation = {
     params: {},
     body: {
-        startDate: Joi.date().optional(),
-        endDate: Joi.date().optional(),
-        deadDate: Joi.date().optional()
+        extensions: Joi.object({
+            startDate: Joi.date().optional(),
```
PanopticaRising, line 106: The problems you caught with us putting moments in the database didn't originate from Joi, right? If we hit these endpoints with moment from the frontend, everything is still fine?

src/features/courses/course-route-validation.ts
```diff
@@ -102,13 +102,22 @@ export const updateCourseTopicValidation = {
...
+            endDate: Joi.date().optional(),
+            deadDate: Joi.date().optional()    
+        }).optional(),
+        studentTopicAssessmentOverride: Joi.object({
+            versionDelay: Joi.number(),
```
PanopticaRising, line 111: Should these be explicitly optional?

src/features/courses/course-route-validation.ts
red-tt-bear: * Correct, the issue @doombot-exe and I saw last night was the backend sending joi along
* I did not see an error on the frontend
* Furthermore a frontend type can't push through to the backend due to serialization and deserialization, if this was a problem joi date would have failed

PanopticaRising (approved):

src/features/courses/course-route-validation.ts
red-tt-bear: They should be, however it doesn't matter for now, I wouldn't hold up the PR for it, once it's merged I'll fix in my current branch

PR #164, opened 2020-10-21
Feature/red 144 exam submit

doombot-exe:
add end exam endpoint

### DIFF
--- a/src/database/models/student-topic-assessment-override.ts
+++ b/src/database/models/student-topic-assessment-override.ts
@@ -36,7 +36,7 @@ export default class StudentTopicAssessmentOverride extends Model implements Stu
     public readonly updatedAt!: Date;
 
     static getOverrides = (obj: StudentTopicAssessmentOverrideOverridesInterface): StudentTopicAssessmentOverrideOverridesInterface => {
-        return _.pick(obj, 'duration', 'maxGradedAttemptsPerVersion', 'maxVersions', 'versionDelay');
+        return _.pick(obj, ['duration', 'maxGradedAttemptsPerVersion', 'maxVersions', 'versionDelay']);
     }
 
     getOverrides = (): StudentTopicAssessmentOverrideOverridesInterface => {
--- a/src/database/models/student-topic-assessment-override.ts
+++ b/src/database/models/student-topic-assessment-override.ts
@@ -1,14 +1,18 @@
 import { Model, DataTypes } from 'sequelize';
 import appSequelize from '../app-sequelize';
+import * as _ from 'lodash';
 
-export interface StudentTopicAssessmentOverrideInterface {
-    id: number;
-    topicAssessmentInfoId: number;
-    userId: number;
+export interface StudentTopicAssessmentOverrideOverridesInterface {
     duration: number | null;
     maxGradedAttemptsPerVersion: number | null;
     maxVersions: number | null;
     versionDelay: number | null;
+}
+
+export interface StudentTopicAssessmentOverrideInterface extends StudentTopicAssessmentOverrideOverridesInterface {
+    id: number;
+    topicAssessmentInfoId: number;
+    userId: number;
     active: boolean;
 }
 
@@ -31,6 +35,13 @@ export default class StudentTopicAssessmentOverride extends Model implements Stu
     public readonly createdAt!: Date;
     public readonly updatedAt!: Date;
 
+    static getOverrides = (obj: StudentTopicAssessmentOverrideOverridesInterface): StudentTopicAssessmentOverrideOverridesInterface => {
+        return _.pick(obj, 'duration', 'maxGradedAttemptsPerVersion', 'maxVersions', 'versionDelay');
+    }
+
+    getOverrides = (): StudentTopicAssessmentOverrideOverridesInterface => {
+        return StudentTopicAssessmentOverride.getOverrides(this);
+    }
     static constraints = {
     }
 
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -16,7 +16,7 @@ import logger from '../../utilities/logger';
 import sequelize = require('sequelize');
 import WrappedError from '../../exceptions/wrapped-error';
 import AlreadyExistsError from '../../exceptions/already-exists-error';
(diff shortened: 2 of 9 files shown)


src/features/courses/course-controller.ts
```diff
@@ -2454,9 +2456,10 @@ class CourseController {
                     await questions.asyncForEach(async (question) => {
                         if (_.isNil(question.grades)) throw new RederlyExtendedError('Impossible! Found an assessment question without a grade.');
                         const version = await courseRepository.getCurrentInstanceForQuestion({questionId: question.id, userId});
-                        question.webworkQuestionPath = version?.webworkQuestionPath ?? question.webworkQuestionPath;
-                        question.grades.randomSeed = version?.randomSeed ?? question.grades.randomSeed;
-                        question.problemNumber = version?.problemNumber ?? question.problemNumber;
+                        if (!_.isNil(version)){
+                            question.grades[0].getVersion(version);
```
red-tt-bear, line 2460: What's the guarantee that a question has grades? We should add length check or null chain this
```
question.grades[0]?.getVersion(version);
```
Also we are not awaiting the response
Lastly we are not using the response (unless it is updating the base model

src/features/courses/course-controller.ts
```diff
@@ -2454,9 +2456,10 @@ class CourseController {
...
+                            question.getVersion(version);
```
PanopticaRising, line 2461: I don't really understand what's happening here.

src/features/courses/course-route.ts
```diff
@@ -312,6 +313,36 @@ router.put('/topic/:id',
...
+
+        if (version.userId !== user.id) {
+            throw new ForbiddenError('You may not end an exam that does not belong to you.');
+        }
+
+        if (version.numAttempts === 0) {
+            throw new ForbiddenError('You cannot end an exam without making at least one attempt.');
+        }
+
+        // if (version.isClean === false) {
```
red-tt-bear, line 337: I would think ending an exam would submit it if there are changes

src/features/courses/course-route.ts
```diff
@@ -312,6 +313,36 @@ router.put('/topic/:id',
...
+        //     // throw error? the student has made changes to their answers, but they have not submitted those changes
+        // }
+
+        version.isClosed = true;
+        version.save();
```
red-tt-bear, line 342: This needs to be awaited

PanopticaRising (approved):

red-tt-bear (changes requested):
Unclear get version calls
Missing awaits
Auto submit closed exam?

src/features/courses/course-controller.ts
doombot-exe: Grades are created on enrollment or on question add, no? We have this check in the data-integrity-checker

You are correct about the updates - I mistakenly thought it was an edit in place.

src/features/courses/course-controller.ts
doombot-exe: this is what Tom asked for -- using the methods on CourseWWTopicQuestion to selectively replace fields according to a StudentGradeInstance...

src/features/courses/course-route.ts
doombot-exe: I have not yet implemented isClean - but yes, this would be ideal

src/features/courses/course-route.ts
doombot-exe: aha

src/features/courses/course-controller.ts
doombot-exe: I'm reverting it anyhow

src/features/courses/course-route.ts
doombot-exe: done

src/features/courses/course-controller.ts
doombot-exe: reverted these because it was too cumbersome to ensure the preservation of includes

doombot-exe:
accounting for overrides on the backend, reverted to the older override method for `getProblems` because of includes and subincludes, and awaited that save promise.

src/features/courses/course-controller.ts
```diff
@@ -2455,7 +2457,7 @@ class CourseController {
                         if (_.isNil(question.grades)) throw new RederlyExtendedError('Impossible! Found an assessment question without a grade.');
                         const version = await courseRepository.getCurrentInstanceForQuestion({questionId: question.id, userId});
                         question.webworkQuestionPath = version?.webworkQuestionPath ?? question.webworkQuestionPath;
-                        question.grades.randomSeed = version?.randomSeed ?? question.grades.randomSeed;
+                        question.grades[0].randomSeed = version?.randomSeed ?? question.grades[0].randomSeed;
```
red-tt-bear, line 2460: Do we need an array length check here?

src/features/courses/course-controller.ts
```diff
@@ -2849,17 +2851,22 @@ class CourseController {
         let message = '';
 
         let topic = await this.getTopicById({ id: topicId }); // includes TopicOverrides
-        const topicInfo = await this.getTopicAssessmentInfoByTopicId({
+        if (!_.isNil(topic.studentTopicOverride)) {
+            topic = topic.getWithOverrides(topic.studentTopicOverride[0]) as CourseTopicContent;
```
red-tt-bear, line 2855: Again do we need an array length check?

src/features/courses/course-controller.ts
```diff
@@ -2849,17 +2851,22 @@ class CourseController {
...
+        }
+
+        let topicInfo = await this.getTopicAssessmentInfoByTopicId({
             topicId: topicId,
             userId: user.id
-        }); // method result is ordered by startDate, includes overrides
+        }); 
+        if (!_.isNil(topicInfo.studentTopicAssessmentOverride)) {
+            topicInfo = topicInfo.getWithOverrides(topicInfo.studentTopicAssessmentOverride[0]) as TopicAssessmentInfo;
```
red-tt-bear, line 2863: Array length check?

red-tt-bear (changes requested):
Array length checks

src/features/courses/course-controller.ts
doombot-exe: grades is an include -- if it's included at all, this will suffice? I think sequelize won't include an empty array here if there's no match.

src/database/models/student-topic-assessment-override.ts
```diff
@@ -31,6 +35,13 @@ export default class StudentTopicAssessmentOverride extends Model implements Stu
     public readonly createdAt!: Date;
     public readonly updatedAt!: Date;
 
+    static getOverrides = (obj: StudentTopicAssessmentOverrideOverridesInterface): StudentTopicAssessmentOverrideOverridesInterface => {
+        return _.pick(obj, 'duration', 'maxGradedAttemptsPerVersion', 'maxVersions', 'versionDelay');
```
PanopticaRising, line 39: https://lodash.com/docs/4.17.15#pick

`pick` looks like it should only take an array.
```suggestion
        return _.pick(obj, ['duration', 'maxGradedAttemptsPerVersion', 'maxVersions', 'versionDelay']);
```

red-tt-bear: I feel like variable arguments is sometimes documented as an array so it might not be wrong, I'll double check

PanopticaRising (approved):

src/database/models/student-topic-assessment-override.ts
red-tt-bear: image
Documentation has it as variable arguments in here
```js
const _ = require('lodash');

const obj = {
    a: 1,
    b: 2,
    c: 3,
    d: 4,
    e: 5
};

console.log(JSON.stringify(_.pick(obj, 'a','c','e')));
```
Results in:
```
{"a":1,"c":3,"e":5}
```

red-tt-bear (approved):

src/features/courses/course-controller.ts
red-tt-bear: It will include an empty array if there is no match

PR #165, opened 2020-10-22
Feature/add renderer preview by question path

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -1216,6 +1216,11 @@ class CourseController {
             calculatedRendererParameters.outputformat = OutputFormat.STATIC;
         }
 
+        // I do not like hardcoding the topicTypeId...
```
PanopticaRising, line 1219: This might've been a merge that should've been deleted

PanopticaRising: @red-tt-bear @doombot-exe Can I get a confirmation before I remove it?

red-tt-bear: Not sure, this was @doombot-exe s code

red-tt-bear: Though if it's coming as an addition I would tend to agree

red-tt-bear (approved):

src/features/courses/course-controller.ts
doombot-exe: I moved it into another block

doombot-exe (changes requested):
remove reassigning calculatedRendererParams (already happens at 1192)

doombot-exe (approved):
seems that's already been removed.

PR #167, opened 2020-10-22
Feature/red 117 auto submit

red-tt-bear:


Dockerfile
```diff
@@ -17,6 +17,8 @@ COPY . ./
 
 #COPY . ./
 RUN npm run build
+# Needs to be before prune since it uses sequelize-cli
+RUN npm run sequelize:built:migrations
```
red-tt-bear, line 21: The only problem I see with this is that in order to build you need knowledge of the db which I don't love
Granted the .env gets copied over right now so you have the knowledge of the database

What I like about this is that it is automatic (thus less prone to human error) and it doesn't require me to manually run anything against prod (right now I have to make sure i'm on the right branch, update to prod creds, run it at the right time, and then make sure to delete prod creds from my env)

doombot-exe (approved):

PanopticaRising (approved):

Jhuanderson-Rederly (approved):

PR #178, opened 2020-10-23
throw errors instead of Boom

doombot-exe:
improved error handling on getting question(s)

### DIFF
--- a/src/features/courses/course-route.ts
+++ b/src/features/courses/course-route.ts
@@ -193,7 +193,7 @@ router.get('/questions',
                 const session = req.session;
                 userId = session.userId;
             } else {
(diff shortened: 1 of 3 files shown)


PanopticaRising (approved):

src/features/courses/course-route.ts
```diff
@@ -193,7 +193,7 @@ router.get('/questions',
                 const session = req.session;
                 userId = session.userId;
             } else {
-                next(Boom.badRequest('userIdInput as a string must be the value `me`'));
+                throw new IllegalArgumentException('userIdInput as a string must be the value `me`');
```
red-tt-bear, line 196: Good defensive code, however this can also be handled at the joi level

PR #180, opened 2020-10-23
Bugfix/red 199 localize next version date

red-tt-bear:


doombot-exe (commented):
this is all terribly clever

doombot-exe (approved):

PR #196, opened 2020-10-26
refetch overrides because sequelize truncated fields

doombot-exe:
it was only a matter of time...
1) don't apply overrides when you have nested sequelize objects that you will later need methods from
2) nested includes can get truncated when the fields get too long


src/features/courses/course-controller.ts
```diff
@@ -2568,7 +2568,7 @@ class CourseController {
                         // TODO: prevent adding problems to an exam once versions have been generated
                         if (_.isNil(version)) {
                             questions.splice(index, 1);
-                            logger.error(`Topic #${topic.id}: Assessment version has a problem (${question.id}) with grade #${question.grades[0]} but no grade instance`);
+                            logger.error(`Topic #${topic.id}: Assessment version has a problem (${question.id}) with grade #${question.grades[0].id} but no grade instance`);
```
red-tt-bear, line 2571: is the array guaranteed to have 1 grade or is our log going to throw a null pointer?
Do we need a question mark after 0

red-tt-bear (approved):

PR #199, opened 2020-10-26
RED-225 assessment workbook proper path & seed

doombot-exe:


src/features/courses/course-controller.ts
```diff
@@ -1220,6 +1220,16 @@ class CourseController {
             if (studentGrade.courseWWTopicQuestionId !== options.questionId) {
                 throw new NotFoundError('The workbook you have requested does not belong to the question provided');
             }
+            // make sure to grab the right path & seed if this is an assessment workbook!
+            if (_.isNil(workbook.studentGradeInstanceId)) {
+                problemSeed = studentGrade.randomSeed;
+                numIncorrect = studentGrade.numAttempts;
+            } else {
+                gradeInstance = await workbook.getStudentGradeInstance();
```
red-tt-bear, line 1228: Should probably include:
```Typescript
{
where: {
active: true
}
}
```

red-tt-bear: This should never be the case, unless somehow we decide it can be in the future, so it's fine for now

red-tt-bear (approved):

src/features/courses/course-repository.ts
```diff
@@ -573,6 +573,17 @@ class CourseRepository {
...
     }
     
+    async getStudentGradeInstance(options: GetStudentGradeInstanceOptions): Promise<StudentGradeInstance> {
+        const result = await StudentGradeInstance.findOne({
+            where: {
+                id: options.id,
+                active:true
+            }
+        });
+        if (_.isNil(result)) throw new NotFoundError(`Requested grade instance #${options.id} not found`);
```
red-tt-bear, line 583: I'm back and forth on this, I feel like it should be the caller's responsibility to null check it
I'm fine either way though

red-tt-bear (approved):

PR #203, opened 2020-10-26
Implemented update-else-create flow for assessments, removed unnecess…

PanopticaRising:
…ary unique constraint.

### DIFF
--- a/src/features/courses/course-repository.ts
+++ b/src/features/courses/course-repository.ts
@@ -288,18 +288,30 @@ class CourseRepository {
                 returning: true,
             });
 
-            if (updates[0] === 1 && !_.isEmpty(options.updates.topicAssessmentInfo)) {
+            if (updates[0] === 1 && 
+                !_.isUndefined(options.updates.topicAssessmentInfo) && 
+                !_.isEmpty(options.updates.topicAssessmentInfo)) {
+
                 const updatedObj: CourseTopicContent = updates[1][0];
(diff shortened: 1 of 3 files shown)


src/features/courses/course-repository.ts
```diff
@@ -288,18 +288,30 @@ class CourseRepository {
                 returning: true,
             });
 
-            if (updates[0] === 1 && !_.isEmpty(options.updates.topicAssessmentInfo)) {
+            if (updates[0] === 1 && 
+                !_.isUndefined(options.updates.topicAssessmentInfo) &&
```
red-tt-bear, line 292: Any reason for undefined check rather than null check?

src/features/courses/course-repository.ts
```diff
@@ -666,19 +678,28 @@ class CourseRepository {
                 logger.error('A single question was expected to update, but multiple update objects are returned.');
             }
 
-            if (updates[0] === 1 && !_.isEmpty(options.updates.courseQuestionAssessmentInfo)) {
+            if (updates[0] === 1 &&
+                !_.isEmpty(options.updates.courseQuestionAssessmentInfo) &&
```
red-tt-bear, line 682: NIT: Order is different from above, I feel like I like nil check first, doesn't matter either way

red-tt-bear (approved):

doombot-exe (approved):

PR #210, opened 2020-10-27
Feature/red 125 attachments

red-tt-bear:


doombot-exe (approved):

src/playground-scheduler-functions.ts
```diff
@@ -0,0 +1,93 @@
+import schedulerHelper, { HttpMethod } from './utilities/scheduler-helper';
```
PanopticaRising, line 1: Should there be a playground directory, rather than just under src?

PanopticaRising (approved):

src/playground-scheduler-functions.ts
red-tt-bear: Not a bad idea, I just moved this out for now
I was thinking we would either restructure or delete the file
it might make sense to put it in a subdirectory and then exclude it from the build?

PR #211, opened 2020-10-27
RED-233 backend

doombot-exe:
added grades endpoint for fetching grades associated to an assessment (this can be moved off the courses/assessment/ route because it equally applies to any topic


src/features/courses/course-controller.ts
```diff
@@ -3271,6 +3272,63 @@ class CourseController {
...
+            });
+
+            if (_.isNil(topic)) {
+                throw new WrappedError(`failed to retrieve topic #${topicId} for grading`);
+            }
+
+            const problems = await CourseWWTopicQuestion.findAll({
+                where: {
+                    courseTopicContentId: topicId,
+                    active: true
```
red-tt-bear, line 3305: Do we need to account for hidden? We have the flag in the db but no way to use it now

red-tt-bear (approved):

src/features/courses/course-controller.ts
```diff
@@ -3271,6 +3272,63 @@ class CourseController {
             return { problemScores: problemScoresReturn, bestVersionScore: bestVersionScoreReturn, bestOverallVersion: bestOverallVersionReturn};
         });
     };
+    async canUserGradeAssessment({
+        user,
+        topicId
+    }: CanUserGradeAssessmentOptions): Promise<boolean> {
+        const topic = await courseRepository.getCourseTopic({ id: topicId });
```
PanopticaRising, line 3279: Should this be 3 separate calls instead of one call with includes? (What is this, GraphQL?)

src/features/courses/course-controller.ts
```diff
@@ -3271,6 +3272,63 @@ class CourseController {
...
+                    id: topicId,
+                    active: true,
+                },
+            });
+
+            if (_.isNil(topic)) {
+                throw new WrappedError(`failed to retrieve topic #${topicId} for grading`);
+            }
+
+            const problems = await CourseWWTopicQuestion.findAll({
```
PanopticaRising, line 3302: Should this just be an includes on the get topic?

src/features/courses/course-route.ts
```diff
@@ -287,6 +287,30 @@ router.put('/topic/:id',
...
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, GradeAssessmentRequest.body, GradeAssessmentRequest.query>, _res: Response, next: NextFunction) => {
+        const params = req.params as GradeAssessmentRequest.params;
+        if (_.isNil(req.session)) {
+            throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+
+        const user = await req.session.getUser();
+        if (await courseController.canUserGradeAssessment({user, topicId: params.id}) === false) {
+            throw new ForbiddenError('You are not allowed to grade this assessment.');
+            return;
```
PanopticaRising, line 304: Unnecessary return after throw

PanopticaRising (approved):
I'm fine with not optimizing our DB calls right now

src/features/courses/course-controller.ts
red-tt-bear: I like it as 3 separate in some regards, I think it makes it easier to pass it in, but I also like it in one because it's potentially more optimized on it's own, I'm back and forth and I think in our stabilization sprint we will need to investigate the performance differences

src/features/courses/course-controller.ts
```diff
@@ -3271,6 +3272,59 @@ class CourseController {
...
+                        model: StudentWorkbook,
+                        as: 'workbooks',
+                        required: false,
+                        where: {
+                            active: true,
+                        }
+                    }]
+                }]
+            });
+            return {problems, topic};
```
red-tt-bear, line 3323: Once you move the includes you can just return the topic 🤣

red-tt-bear (approved):

PR #213, opened 2020-10-27
RED-125 Added list attachment route and logic

red-tt-bear:


src/features/courses/course-route-validation.ts
```diff
@@ -458,3 +458,12 @@ export const postAttachmentValidation = {
         studentGradeInstanceId: Joi.number().optional()
     },
 };
+
+export const listAttachmentsValidation = {
+    params: {},
+    query: {
+        studentGradeId: Joi.number().optional(),
```
PanopticaRising, line 465: You can put XOR logic here with refs if you want to get fancy.

PanopticaRising: https://joi.dev/api/?v=17.3.0#alternatives

PanopticaRising (approved):

src/features/courses/course-route-validation.ts
red-tt-bear: I wonder how that plays with extract types, I think I would like it in both places
we'll revisit with the refactor

PR #214, opened 2020-10-28
RED-125 Added delete route and logic for attachments

red-tt-bear:


doombot-exe (approved):

src/features/courses/course-controller.ts
```diff
@@ -3275,16 +3303,38 @@ class CourseController {
...
+        studentWorkbookId
     }: CreateAttachmentOptions): Promise<ProblemAttachment> {
         return useDatabaseTransaction(async (): Promise<ProblemAttachment> => {
-            if (_.isNil(studentGradeId) === _.isNil(studentGradeInstanceId)) {
-                throw new IllegalArgumentException('Student grade XOR student grade instance id must be supplied');
+            const filterCount = [
+                studentGradeId,
+                studentGradeInstanceId,
+                studentWorkbookId,
+            ].reduce((accumulator, val) => (accumulator || 0) + (!_.isNil(val) && 1 || 0), 0);
```
PanopticaRising, line 3314: Is this equivalent? You're just counting the non-nill values in the array?
```suggestion
            ].reduce((accumulator, val) => accumulator + (!_.isNil(val) ? 1 : 0), 0);
```

red-tt-bear: I kept this consistent with how I've done it in the past, I'll revisit later at all three places at the same time

I think accumulator might be nil the first iteration (or maybe I didn't have the second paramater when I first wrote it)
Either way I don't think it matters and it's just more defensive

The second part is semantics, I'm saying nill & 1 will result in nill otherwise 1 and then defaulting to 0
You are doing the same thing with a ternary operator, I'm fine making that change since I agree it's a little easier but last time it got through so I would want to clean it all up together

PanopticaRising (approved):

src/features/courses/course-controller.ts
PanopticaRising: The second operand will guarantee that accum doesn't start null. Mixing && and || is against the frontend ESLint rules (one of the few cases I think is actually a valid "confusing semantic")

PR #215, opened 2020-10-28
Added enpoint to allow students to email the professor of a course.

PanopticaRising:
Also updated the reply-to header in the mailer so professors can easily reply to the student.

### DIFF
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -3578,7 +3578,7 @@ class CourseController {
 Hello Professor ${course.instructor.lastName},
 
     Your student ${options.student.firstName} ${options.student.lastName} is asking for help with
-Problem ${question.problemNumber+1} in the Topic ${topic.name}.
+Problem ${question.problemNumber} in the Topic ${topic.name}.
 
 Here is the message that was sent:
 
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -3543,30 +3543,42 @@ class CourseController {
             }],
         });
 
-        const gradePromise = StudentGrade.findOne({
-            where: {
-                userId: options.student.id,
-                courseWWTopicQuestionId: options.question.id,
-            },
-            attributes: ['randomSeed']
+        const topicPromise = CourseTopicContent.findOne({
+            attributes: ['id', 'name'],
+            where: {},
+            include: [{
+                model: CourseWWTopicQuestion,
+                as: 'questions',
+                attributes: ['problemNumber'],
+                required: true,
+                where: {
+                    id: options.question.id,
+                    active: true,
+                }
+            }]
         });
 
-        const [course, grade] = await Promise.all([coursePromise, gradePromise]);
+        const [course, topic] = await Promise.all([coursePromise, topicPromise]);
 
         if (_.isNil(course) || _.isNil(course.instructor)) {
-            throw new WrappedError('Could not find an instructor for this course.');
+            throw new RederlyError('Could not find an instructor for this course.');
         }
 
-        if (_.isNil(grade)) {
-            throw new WrappedError('Could not find a grade associated with the problem this student is trying to ask about.');
+        if (_.isNil(topic)) {
+            throw new RederlyError('Could not find a topic associated with the problem this student is trying to ask about.');
+        }
+        
+        const question = topic.questions?.[0];
+        
+        if (_.isNil(question)) {
+            throw new RederlyError('Could not find the question associated with the problem/topic this student is trying to ask about.');
         }
 
         const poorMansTemplate = `
 Hello Professor ${course.instructor.lastName},
 
     Your student ${options.student.firstName} ${options.student.lastName} is asking for help with
-Problem ${options.question.problemNumber+1} (ID#${options.question.id}) in the Topic ${options.topic.name} (ID#${options.topic.id})
-The WeBWorK path for this problem is ${options.question.webworkQuestionPath} and the seed value is ${grade.randomSeed}.
+Problem ${question.problemNumber+1} in the Topic ${topic.name}.
 
 Here is the message that was sent:
 
@@ -3578,8 +3590,8 @@ You should be able to reply to the student's email address (${options.student.pr
         return emailHelper.sendEmail({
             content: poorMansTemplate,
             email: course.instructor.preferredEmail,
-            subject: `${options.student.firstName} - Topic ${options.topic.id} - Question ${options.question.id}`,
-            from: options.student.preferredEmail,
+            subject: `${options.student.firstName} - Topic ${topic.id} - Question ${options.question.id}`,
+            replyTo: options.student.preferredEmail,
         });
     }
 }
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -16,7 +16,7 @@ import logger from '../../utilities/logger';
 import sequelize = require('sequelize');
 import WrappedError from '../../exceptions/wrapped-error';
 import AlreadyExistsError from '../../exceptions/already-exists-error';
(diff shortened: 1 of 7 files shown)


src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+                userId: options.student.id,
+                courseWWTopicQuestionId: options.question.id,
+            },
+            attributes: ['randomSeed']
+        });
+
+        const [course, grade] = await Promise.all([coursePromise, gradePromise]);
+
+        if (_.isNil(course) || _.isNil(course.instructor)) {
+            throw new WrappedError('Could not find an instructor for this course.');
```
red-tt-bear, line 3395: I've been bad about this too but this should be RederlyError

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+        }
+
+        if (_.isNil(grade)) {
+            throw new WrappedError('Could not find a grade associated with the problem this student is trying to ask about.');
+        }
+
+        const poorMansTemplate = `
```
red-tt-bear, line 3402: 🤣

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+Hello Professor ${course.instructor.lastName},
+
+    Your student ${options.student.firstName} ${options.student.lastName} is asking for help with
+Problem ${options.question.problemNumber+1} (ID#${options.question.id}) in the Topic ${options.topic.name} (ID#${options.topic.id})
```
red-tt-bear, line 3406: Why are we incrementing problemNumber by 1?

Also I still don't like the id int the request, it's great for reordering but i'm skeptical on putting these numbers in there

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+The WeBWorK path for this problem is ${options.question.webworkQuestionPath} and the seed value is ${grade.randomSeed}.
```
red-tt-bear, line 3407: We should not put the webwork path or seed in here, too much info and if we add a reply to student from this email (which I think you did) we will be exposing bad things!

Furthermore it is like putting a jira ticket in a message, no one knows what it means until they look at it anyway therefore it's not even particularly useful

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+${options.content}
+
+You should be able to reply to the student's email address (${options.student.preferredEmail}) by replying to this message.
+`;
+
+        return emailHelper.sendEmail({
+            content: poorMansTemplate,
+            email: course.instructor.preferredEmail,
+            subject: `${options.student.firstName} - Topic ${options.topic.id} - Question ${options.question.id}`,
+            from: options.student.preferredEmail,
```
red-tt-bear, line 3420: Do we have issues with it detecting as spam? I feel like including a from with an unverified email typically results in instant spam

src/features/courses/course-route-validation.ts
```diff
@@ -467,3 +467,22 @@ export const listAttachmentsValidation = {
...
+    query: {},
+    body: {
+        content: Joi.string().required(),
+        topic: Joi.object({
+            id: Joi.number().required(),
+            name: Joi.string().required(),
+        }).required(),
+        question: Joi.object({
+            id: Joi.number().required(),
+            webworkQuestionPath: Joi.string().required(),
```
red-tt-bear, line 484: We should remove this

red-tt-bear: Problem number as well, we should be getting this from a fetched database object, otherwise a malicious user could spoof easily

Yes we could add permission checks but then we will be running the same queries anyway

src/features/courses/course-controller.ts
red-tt-bear: I felt like their was probably a distinction and now I see it: `...(options.from ? {headers: {'Reply-To': options.from }} : undefined)`

I would assume this is good but you should try from other mail platforms other than gmail

Either way I think it is fine for now

red-tt-bear (changes requested):
Let's get rid of the sensitive webwork information from the email that can be replied to

src/features/courses/course-controller.ts
doombot-exe: reply-to is the way to go here

src/features/courses/course-controller.ts
PanopticaRising: This is user-facing, so it needs to be offset by one.

It doesn't hurt anything to have more information right now. If the professor can see the ID numbers anywhere, then they can use this. I would only remove it if the information was misleading, like that the ID messages don't match what's on the Settings page.

src/features/courses/course-controller.ts
PanopticaRising: I guess I'll remove this and we'll just have to remember to update it with @doombot-exe 's stuff after the fact.

src/features/courses/course-controller.ts
PanopticaRising: sendGrid actually blocks you from using `from` from an unverified domain. ReplyTo works, I just didn't change the name in the option.

src/features/courses/course-route-validation.ts
PanopticaRising: Where does the problemNumber change when randomizeOrder is true?

PanopticaRising: I guess the problemId doesn't change, and the user doesn't see that, so it should work regardless.

PanopticaRising: I wouldn't really think there's much for a malicious user to spoof here. We could lock it down and always do the database calls, but the important information- the user initiating the request- is gotten from the session.

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,63 @@ class CourseController {
...
+                as: 'instructor',
+                attributes: ['id', 'firstName', 'lastName', 'email', 'preferredEmail'],
+                where: {
+                    active: true
+                },
+                required: true,
+            }],
+        });
+
+        const gradePromise = StudentGrade.findOne({
```
PanopticaRising, line 3384: @doombot-exe @red-tt-bear Do I need to also be getting the gradeInstance here, if there's one applicable? I don't remember if the randomSeed bubbles up.

PanopticaRising: Just kidding, I'm only getting the seed from this so I don't need it at all 🤦

src/features/courses/course-controller.ts
red-tt-bear: problem number should be misleading if it's not there is probably a bigger issue here

red-tt-bear: As @doombot-exe pointed out yesterday too much information can be overbearing, the only advantage to showing the id here is if the problem has been reordered

src/features/courses/course-controller.ts
red-tt-bear: Right, that's why I was misled, look good though

src/features/courses/course-route-validation.ts
red-tt-bear: I mean even if it's not a spoof, I think the database queries have to occur anyway, that will always be 100% up to date as well as avoid any bugs on the frontend (like state management bugs)

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,75 @@ class CourseController {
...
+
+    Your student ${options.student.firstName} ${options.student.lastName} is asking for help with
+Problem ${question.problemNumber+1} in the Topic ${topic.name}.
+
+Here is the message that was sent:
+
+${options.content}
+
+You should be able to reply to the student's email address (${options.student.preferredEmail}) by replying to this message.
+`;
```
red-tt-bear, line 3426: Shouldn't we have a signature?

src/features/courses/course-controller.ts
```diff
@@ -3362,6 +3363,75 @@ class CourseController {
...
+        
+        if (_.isNil(question)) {
+            throw new RederlyError('Could not find the question associated with the problem/topic this student is trying to ask about.');
+        }
+
+        const poorMansTemplate = `
+Hello Professor ${course.instructor.lastName},
+
+    Your student ${options.student.firstName} ${options.student.lastName} is asking for help with
+Problem ${question.problemNumber+1} in the Topic ${topic.name}.
```
red-tt-bear, line 3419: Still not convinced on problemNumber +1, we will have to talk during scrum

src/features/courses/course-route.ts
```diff
@@ -941,6 +941,31 @@ router.get('/topics',
...
     }));
 
+router.post('/:id/email',
+    authenticationMiddleware,
+    validate(emailProfValidation),
+    // This is due to a typescript issue where the type mismatches extractMap
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, EmailProfRequest.body, EmailProfRequest.query>, res: Response, next: NextFunction) => {
+        const params: EmailProfRequest.params = req.params;
+        const body = req.body as EmailProfRequest.body;
```
red-tt-bear, line 951: You don't need this...

red-tt-bear (approved):

src/features/courses/course-controller.ts
red-tt-bear: This is tricky though because you need the grade instance for the student's problem number...
The problem number for the professor and the user is not necessarily the same
Good thing you included the id 😅 
Once we get nice routing we could not give any of that information and say:

Your student is having an issue with a problem. Click this link for more information

One other thing that could be looked at as a pro or a con by not giving extra information in the email is that it never gets stale, reordering the question won't matter (though I don't think this is a big deal)

src/features/courses/course-controller.ts
PanopticaRising: Whoops, you're right, I must've gotten mixed up, I thought problemNumber was 0-indexed.

src/features/courses/course-controller.ts
PanopticaRising: A signature for what?

src/features/courses/course-route.ts
PanopticaRising: Look man, I was having some typing problems!

src/features/courses/course-controller.ts
red-tt-bear:
```
All the best,
The Rederly Team
```

PR #217, opened 2020-10-29
The S3-Lambda API has changed. This update matches the expected API r…

PanopticaRising:
…esponse.

### DIFF
--- a/src/utilities/attachments-helper.ts
+++ b/src/utilities/attachments-helper.ts
@@ -9,7 +9,9 @@ interface AttachmentHelperOptions {
 };
 
 interface RequestPresignedURLResponse {
-    upload_url: string;
+    // TODO: URL type?
+    uploadURL: string;
+    photoFilename: string;
 }
 
 interface GetPresignedURLResponse {
@@ -48,8 +50,8 @@ class AttachmentHelper {
     getNewPresignedURL = async (): Promise<GetPresignedURLResponse> => {
         const result = await this.requestNewPresignedURL();
         return {
-            uploadURL: result.data.upload_url,
-            cloudFilename: this.parseTokenFromPresignedURL(result.data.upload_url)
+            uploadURL: result.data.uploadURL,
+            cloudFilename: result.data.photoFilename,
         };
     }
 


src/utilities/attachments-helper.ts
```diff
@@ -9,7 +9,9 @@ interface AttachmentHelperOptions {
 };
 
 interface RequestPresignedURLResponse {
-    upload_url: string;
+    // TODO: URL type?
+    uploadURL: string;
+    photoFilename: string;
```
red-tt-bear, line 14: has nothing to do with this code but I don't like the name photoFilename unless it will always be a photo
I thought the lambda generated pdfs?

src/utilities/attachments-helper.ts
```diff
@@ -9,7 +9,9 @@ interface AttachmentHelperOptions {
 };
 
 interface RequestPresignedURLResponse {
-    upload_url: string;
+    // TODO: URL type?
```
red-tt-bear, line 12: We shouldn't be doing any backend logic with the url since it is now passing the filename separate so it doesn't add any value

We could validate that it is a valid url but what if in the future it was a relative url (like we used the same cloudfront base for our uploads

I can see either the backend or frontend validating the urls integrity (especially now that the frontend reports errors back)

red-tt-bear (approved):

src/utilities/attachments-helper.ts
PanopticaRising: Preventative typing doesn't have to be used immediately for it to add value. But you did use extra logic with when you were parsing out the filename (though this typing wouldn't strictly help with that either).

In general, I think we prefer to err on the side of specificity with our typings. That's one of the benefits, having to explicitly change the types when you're changing the logic.

src/utilities/attachments-helper.ts
PanopticaRising: We should get the lambda code into a repository. This was due to reverting back to an older Node.JS script, rather than the newer Python script that you had originally integrated with.

PR #225, opened 2020-10-31
Feature/red 232 integrated editor

red-tt-bear:


src/utilities/renderer-helper.ts
```diff
@@ -131,29 +138,29 @@ export const rendererResponseValidationScheme = Joi.object({
     })).required(),
     debug: Joi.object({
         // TODO are these required or optional
-        debug: Joi.array().items(Joi.string()).required(),
-        internal: Joi.array().items(Joi.string()).required(),
-        perl_warn: Joi.string().allow('').required(),
-        pg_warn: Joi.array().items(Joi.string()).required(),
-        render_warn: Joi.array().items(Joi.string()).optional(), // THIS FIELD IS NEW, replace with required 
+        // debug: Joi.array().items(Joi.string()).required(),
```
red-tt-bear, line 141: @doombot-exe we should sync on this, I had issues where it said it wasn't an array when I used the preview with the source

(but I don't think this happened with the renderer helper)

src/features/courses/course-route.ts
```diff
@@ -1117,4 +1118,67 @@ router.delete('/attachments/:id',
...
+        if (_.isNil(req.session)) {
+            throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+
+        const basePath = `private/my/${req.session.userId}`;
+        const result = await rendererHelper.catalog({
+            basePath: basePath,
+            // TODO what should the depth be?
+            // -1 did not do all
+            maxDepth: 100,
```
red-tt-bear, line 1177: I thought -1 was supposed to do the whole catalog

drdrew42: I don't think I ever built that in. There is no "unlimited" search.

PanopticaRising: Should maxDepth be passed in from the UI?

PanopticaRising (approved):

src/features/courses/course-route.ts
red-tt-bear: In the future yes, the base path prefix is also hardcoded here
I feel like this makes it a little more secure in the meantime

PR #229, opened 2020-11-01
fix early access overrides to all topics

doombot-exe:


src/features/courses/course-repository.ts
```diff
@@ -544,6 +544,21 @@ class CourseRepository {
...
+    async getStudentTopicOverride(options: {userId: number; topicId: number}): Promise<StudentTopicOverride | null> {
+        try {
+            const override = await StudentTopicOverride.findOne({
+                where: {
+                    userId: options.userId,
+                    courseTopicContentId: options.topicId,
+                    active: true
+                }
+            });
+            return override; // null is fine, there might not *be* any override for this combo...
```
red-tt-bear, line 556: I think one cool way to do this would to have a `throw` option in the settings that defaults to true
This way we can default it and if we know it can be null we can set it to false

The problem is if it can be null we will have to add an additional null check, might make sense to just always put that one layer up

src/features/courses/course-repository.ts
```diff
@@ -544,6 +544,21 @@ class CourseRepository {
         }
     }
 
+    async getStudentTopicOverride(options: {userId: number; topicId: number}): Promise<StudentTopicOverride | null> {
```
red-tt-bear, line 547: We need to clean up inline types in the future

src/features/courses/course-repository.ts
```diff
@@ -544,6 +544,21 @@ class CourseRepository {
...
+            const override = await StudentTopicOverride.findOne({
+                where: {
+                    userId: options.userId,
+                    courseTopicContentId: options.topicId,
+                    active: true
+                }
+            });
+            return override; // null is fine, there might not *be* any override for this combo...
+        } catch (e) {
+            throw new WrappedError(`Failed while finding student topic override for user ${options.userId} and topic ${options.topicId}.`);
```
red-tt-bear, line 558: Should be RederlyError

red-tt-bear: Nope I lied, it should be wrapped error, but you should pass e along with it

src/features/courses/course-controller.ts
```diff
@@ -3027,15 +3027,21 @@ class CourseController {
...
+        if (!_.isNil(topicOverride)) {
+            topic = topic.getWithOverrides(topicOverride) as CourseTopicContent;
+        }
+
+        // applies to all topics - not just homeworks...
+        if (topic.startDate.toMoment().isAfter(moment())) {
+            message = `${topic.name} hasn't started yet.`;
+            return { userCanViewQuestion: false, message };
+        } else {
+            if (topic.topicTypeId === 1) return { userCanViewQuestion: true, message };
```
red-tt-bear, line 3041: enums

src/features/courses/course-controller.ts
```diff
@@ -3027,15 +3027,21 @@ class CourseController {
...
-                message = `${topic.name} hasn't started yet.`;
-                return { userCanViewQuestion: false, message };
-            } else {
-                return { userCanViewQuestion: true, message };
-            }
-        } else if (topic.topicTypeId === 2) {
+        let topic = await question.getTopic();
+        const topicOverride = await courseRepository.getStudentTopicOverride({userId: user.id, topicId: topic.id});
+        if (!_.isNil(topicOverride)) {
+            topic = topic.getWithOverrides(topicOverride) as CourseTopicContent;
```
red-tt-bear, line 3033: This is casting to an invalid type, the typing will say you have functions and attributes that will fail

i.e. if you try to do topic.getQuestions typescript will say ok and react will crash

red-tt-bear: We should make the type of topic be CourseTopicContentInterface

red-tt-bear (changes requested):
The only thing blocking me is the type that can lead to extreme downstream errors

src/features/courses/course-controller.ts
red-tt-bear: const dbTopic = await question.getTopic();
let topic: CourseTopicContentInterface = dbTopic;
topic = dbTopic.getWithOverrides(topicOverride);

src/features/courses/course-repository.ts
doombot-exe: I think we reconfig our get<object> calls to automatically handle overrides where appropriate, or incorporate a withOverride: boolean

PanopticaRising (approved):

src/features/courses/course-controller.ts
```diff
@@ -3027,15 +3027,19 @@ class CourseController {
...
-            if (topic.startDate.toMoment().isAfter(moment())) {
-                message = `${topic.name} hasn't started yet.`;
-                return { userCanViewQuestion: false, message };
-            } else {
-                return { userCanViewQuestion: true, message };
-            }
-        } else if (topic.topicTypeId === 2) {
+        const dbTopic = await question.getTopic();
+        const topicOverride = await courseRepository.getStudentTopicOverride({userId: user.id, topicId: dbTopic.id});
+        const topic: CourseTopicContentInterface = (_.isNil(topicOverride)) ? dbTopic : dbTopic.getWithOverrides(topicOverride);
```
red-tt-bear, line 3032: We could add the nil check in the getWithOverrides, null overrides should return the original

src/features/courses/course-repository.ts
red-tt-bear: Yeah that's on option, my only problem with doing at a super high level is that we might want to do something with the original data

Also once we apply the overrides we modify the model so it becomes in possible to pass it around and use the mixins

red-tt-bear (approved):

PR #253, opened 2020-11-10
Hotfix/apply override solution date

red-tt-bear:


src/extensions/array-extension.ts
```diff
@@ -10,3 +11,11 @@ Array.prototype.asyncForEach = function <T>(callbackfn: (value: T, index: number
     });
     return Promise.all(promises);
 };
+
+Object.defineProperty(Array.prototype, 'first', {
+    get(this: Array<unknown>) {
+        return this[0];
```
PanopticaRising, line 17: Should this do some safety checking, like, returning null if len = 0?

PanopticaRising (approved):

src/extensions/array-extension.ts
red-tt-bear: ```[][0]``` returns undefined, no need to check as long as long as the array is defined

red-tt-bear: The interface says it can be undefined

PR #256, opened 2020-11-10
check enrollment during canUserStartNewVersion

doombot-exe:


src/features/courses/course-controller.ts
```diff
@@ -2933,7 +2933,9 @@ class CourseController {
                         active: true
                     }
                 });
-                if (_.isNil(questionGrade)) throw new WrappedError(`Question id: ${question.id} has no corresponding grade.`);
+                if (_.isNil(questionGrade)) {
+                    throw new WrappedError(`Question id: ${question.id} has no corresponding grade.`);
```
red-tt-bear, line 2937: I see it was already there but this should be RederlyError

src/features/courses/course-controller.ts
```diff
@@ -3083,6 +3085,24 @@ class CourseController {
         if (!_.isNil(topic.studentTopicOverride) && topic.studentTopicOverride.length > 0) {
             topic = topic.getWithOverrides(topic.studentTopicOverride[0]) as CourseTopicContent;
         }
+        const unit = await topic.getUnit();
+        const course = await unit.getCourse();
+        const enroll = await course.getEnrolledStudents({
+            where: {
```
red-tt-bear, line 3091: Missing where clause params that we need
* dropDate is not null
* active true

src/features/courses/course-controller.ts
```diff
@@ -3083,6 +3085,24 @@ class CourseController {
...
+                userId: user.id,
+            }
+        });
+        if (_.isEmpty(enroll)) {
+            logger.warn(`User #${user.id} is not enrolled in course for topic #${topicId}.`);
```
red-tt-bear, line 3096: I don't think this warning should be here
This is not an error in anyway on the backend

The frontend should report the error since it got to a place it shouldn't have

red-tt-bear (changes requested):
Please fix get enrollments to have all appropriate where clauses

src/features/courses/course-controller.ts
doombot-exe: I think you mean dropDate _is_ null

PanopticaRising (approved):

src/features/courses/course-controller.ts
red-tt-bear: Yup, my bad

red-tt-bear (approved):

PR #257, opened 2020-11-11
Added code to list curriculum that filters out courses that do no mat…

PanopticaRising:
…ch the user's university OR is a rederly public course.

### DIFF
--- a/src/database/models/user.ts
+++ b/src/database/models/user.ts
@@ -26,7 +26,7 @@ export default class User extends Model {
 
   public courseEnrollments?: StudentEnrollment[]
 
-  public getUniversity!: HasOneGetAssociationMixin<University>;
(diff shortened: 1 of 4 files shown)


doombot-exe (approved):
:party-wizard:

src/database/models/user.ts
```diff
@@ -56,9 +56,9 @@ export default class User extends Model {
       as: 'role'
     });
 
-    User.belongsTo(University, {
-      foreignKey: 'universityId',
-      targetKey: 'id',
+    User.hasOne(University, {
```
red-tt-bear, line 59: I don't think this is doing what you expect it to do
I think this is adding an extra foreign key on the university

A user does belong to a university so I don't know why this would change

red-tt-bear (approved):

PR #264, opened 2020-11-17
Feature/red 282 grading page live look

doombot-exe:


src/features/courses/course-controller.ts
```diff
@@ -1217,10 +1217,37 @@ class CourseController {
...
+                    gradeInstance = await courseRepository.getCurrentInstanceForQuestion({
+                        questionId: options.questionId,
+                        userId: options.userId
+                    }); 
+                } else {
+                    // student grade instances do not have a direct reference to userId
+                    // we have to pass through the studentGrade to get there
+                    const questionGrade = await courseQuestion.getGrades({
+                        where: {
+                            userId: options.userId
```
red-tt-bear, line 1230: Should always have active true, though there is no way to set them as inactive right now

src/features/courses/course-controller.ts
```diff
@@ -1217,10 +1217,37 @@ class CourseController {
...
-                }); 
+                if (_.isNil(options.studentTopicAssessmentInfoId)) {
+                    gradeInstance = await courseRepository.getCurrentInstanceForQuestion({
+                        questionId: options.questionId,
+                        userId: options.userId
+                    }); 
+                } else {
+                    // student grade instances do not have a direct reference to userId
+                    // we have to pass through the studentGrade to get there
+                    const questionGrade = await courseQuestion.getGrades({
```
red-tt-bear, line 1228: variable name should be plural

src/features/courses/course-controller.ts
```diff
@@ -1217,10 +1217,37 @@ class CourseController {
...
+                            where: {
+                                studentTopicAssessmentInfoId: options.studentTopicAssessmentInfoId,
+                                studentGradeId: questionGrade[0].id,
+                                active: true
+                            }
+                        });
+                        if (!_.isNil(requestedGradeInstance)) {
+                            gradeInstance = requestedGradeInstance;
+                        } else {
+                            throw new IllegalArgumentException('The version you requested does not correspond to the requested user.');
```
red-tt-bear, line 1245: Couldn't this error message be misleading? If the student gradeId didn't exist?
Though granted I guess the error message still stands 🤷‍♂️

src/features/courses/course-controller.ts
```diff
@@ -2224,6 +2251,49 @@ class CourseController {
...
+                    courseWWTopicQuestionId: questionId,
+                    userId,
+                    active: true,
+                },
+                include
+            });
+            if (!_.isNil(grade)) {
+                return grade;
+            } else {
+                throw new Error('Did not find a student grade.');
```
red-tt-bear, line 2290: This should be rederly error, it helps in debugging and handling to know that the error came from within the application

Furthermore shouldn't this be an illegal argument except because this can be exercised by the API and is not an unhandled exception

src/features/courses/course-controller.ts
```diff
@@ -2224,6 +2251,49 @@ class CourseController {
...
+            }
+        } catch (e) {
+            throw new RederlyError(`Unable to get grade for user #${userId} and question #${questionId}`);
```
red-tt-bear, line 2293: Why are we try catching and then throwing again? Unless there is a specific error to send it makes sense to let it boil up so it can be handled accordingly

src/features/courses/course-route.ts
```diff
@@ -603,6 +603,57 @@ router.post('/question',
         }
     }));
 
+router.get('/question/:id/raw',
```
red-tt-bear, line 606: What is meant by get a raw question? I think we need a method comment on this one

src/features/courses/course-route.ts
```diff
@@ -603,6 +603,57 @@ router.post('/question',
...
+    authenticationMiddleware,
+    validate(getQuestionRawValidation),
+    // This is a typescript workaround since it tries to use the type extractMap
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, GetQuestionRawRequest.body, unknown>, _res: Response, next: NextFunction) => {
+        if (_.isNil(req.session)) {
+            throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
```
red-tt-bear, line 613: This is probably how it was because the Exceptions have been built out over time, but I think this should be RederlyError

src/features/courses/course-route.ts
```diff
@@ -603,6 +603,57 @@ router.post('/question',
...
+        }
+
+        const { id: questionId } = req.params as GetQuestionRawRequest.params;
+        const { userId } = req.query as GetQuestionRawRequest.query;
+        
+        try {
```
red-tt-bear, line 619: Async handler was added after the fact so we have been moving away from doing try catches since it automatically calls next

src/features/courses/course-route.ts
```diff
@@ -613,40 +664,32 @@ router.get('/question/:id',
             throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
         }
 
-        const session = req.session;
-        const user = await req.session.getUser();
+        const requestingUser = await req.session.getUser();
```
red-tt-bear, line 667: Shouldn't need it but active check to be safe?

red-tt-bear (changes requested):
Error handling cleanup

src/features/courses/course-route.ts
doombot-exe: we had branching logic in get question - this is just getting the question fields from the db table.

src/features/courses/course-controller.ts
doombot-exe: this is only ever going to get one grade entry at most.

src/features/courses/course-controller.ts
```diff
@@ -1217,10 +1217,38 @@ class CourseController {
...
+                            userId: options.userId,
+                            active: true,
+                        }
+                    });
+
+                    if (!_.isEmpty(questionGrade)){
+                        const requestedGradeInstance = await StudentGradeInstance.findOne({
+                            where: {
+                                studentTopicAssessmentInfoId: options.studentTopicAssessmentInfoId,
+                                studentGradeId: questionGrade[0].id,
```
PanopticaRising, line 1239: Use `.first`

src/features/courses/course-controller.ts
```diff
@@ -1269,8 +1297,10 @@ class CourseController {
         }
 
         let showCorrectAnswers = false;
+        let answersSubmitted: number | undefined;
         if (options.role === Role.PROFESSOR && !_.isNil(workbook)) {
             showCorrectAnswers = true;
+            answersSubmitted = 1;
```
red-tt-bear, line 1303: Is this a new parameter, even though Perl doesn't type booleans (which makes me 👿 ) I have been typing them as booleans and converting them at the renderer level

When I first looked at this I thought that field represented a count and couldn't understand why you were resetting it)

src/features/courses/course-controller.ts
red-tt-bear: Does it make sense to throw a wrapped error and include it as a parameter?

src/features/courses/course-route-validation.ts
```diff
@@ -269,6 +269,28 @@ export const extendCourseTopicQuestionValidation = {
     },
 };
 
+export const getQuestionGradeValidation = {
+    params: {
+        id: Joi.number().required(),
+    },
+    query: {
+        userId: Joi.number().required(),
+        topicAssessmentInfoId: Joi.number().optional(),
```
PanopticaRising, line 278: Is this option used?

PanopticaRising (approved):

red-tt-bear (approved):

PR #267, opened 2020-11-19
Feature/red 301 user UUID

red-tt-bear:
Make sure to review my other pr first otherwise you will be double reviewing changes

### DIFF
--- a/src/database/migrations/20201118202558-user-uuid.ts
+++ b/src/database/migrations/20201118202558-user-uuid.ts
@@ -1,0 +1,36 @@
+import { QueryInterface, DataTypes } from 'sequelize';
+
+export default {
+    up: async (queryInterface: QueryInterface): Promise<void> => {
+        // Transactions are automatically use because a namespace is injected into sequelize when fetching configurations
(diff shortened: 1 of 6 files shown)


src/database/migrations/20201118202558-user-uuid.ts
```diff
@@ -0,0 +1,53 @@
...
+            );
+
+            await queryInterface.changeColumn(
+                'users',
+                'user_uuid',
+                {
+                    type: DataTypes.UUID,
+                    allowNull: false,
+                    unique: true,
+                    defaultValue: DataTypes.UUIDV4
```
PanopticaRising, line 25: Is this a value or a type?

PanopticaRising (approved):

src/database/migrations/20201118202558-user-uuid.ts
red-tt-bear: This is an enum that represents a function that returns a value: like DataTypes.NOW

UUID is the type but it can be a v1 or a v4

PR #274, opened 2020-11-23
Feature/email templates library

PanopticaRising:


emails/verification/html.pug
```diff
@@ -0,0 +1,318 @@
...
+  meta(name='viewport' content='width=device-width')
+  // Forcing initial-scale shouldn't be necessary
+  meta(http-equiv='X-UA-Compatible' content='IE=edge')
+  // Use the latest (edge) version of IE rendering engine
+  meta(name='x-apple-disable-message-reformatting')
+  // Disable auto-scale in iOS 10 Mail entirely
+  title
+  // The title tag shows in email notifications, like Android 4.4.
+  link(href='https://fonts.googleapis.com/css?family=Poppins:200,300,400,500,600,700' rel='stylesheet')
+  link(rel="preconnect" href="https://fonts.gstatic.com")
```
PanopticaRising, line 14: Considering using `preload` instead: https://web.dev/uses-rel-preconnect/

package.json
```diff
@@ -82,7 +87,7 @@
     "cli": "ts-node ./src/cli.ts",
     "cli:debug": "node --inspect=5859 -r ts-node/register ./src/cli.ts",
     "cli:debug:watch": "nodemon -x npm run cli:debug",
-    "play": "ts-node src/playground/playground.ts",
+    "play": "ts-node src/playground/playground-pug.ts",
```
PanopticaRising, line 90: Revert

package.json
```diff
@@ -113,7 +118,7 @@
     "url": "git+https://github.com/rederly/backend.git"
   },
   "author": "Rederly",
-  "license": "GPL-3.0 License",
+  "license": "GPL-3.0-or-later",
```
PanopticaRising, line 121: Confirm or use a link to LICENSE.md

src/features/users/user-controller.ts
```diff
@@ -357,15 +358,18 @@ class UserController {
             await user.save();
         }
 
+
+        // const verificationEmailTemplateFunction = pug.compileFile('src/email-templates/verification.pug');
```
PanopticaRising, line 362: Remove

package.json
PanopticaRising: Confirmed with text diff.

package.json
```diff
@@ -76,7 +81,7 @@
   "scripts": {
     "test": "jest",
     "start": "ts-node src/index.ts",
-    "start:watch": "nodemon -x npm start",
+    "start:watch": "npx nodemon -x npm start",
```
red-tt-bear, line 84: What is the benefit of using npx?

The reason I use it directly is I like to have it locked down by the package json, when you use npx you don't know what version you're going to get

src/playground/playground-pug.ts
```diff
@@ -0,0 +1,25 @@
...
+    logger.info(`${enabledMarker} EMAIL ENABLED ${enabledMarker}`);
+} else {
+    logger.info(`${disabledMarker} EMAIL DISABLED ${disabledMarker}`);
+}
+
+// Pug playground doesn't specifically need the database.
+// import { sync } from '../database';
+
+
+(async (): Promise<void> => {
```
red-tt-bear, line 20: This is a little different than how I normally do the playgrounds, if I'm going to be totally honest I usually do all the work in the main playground file, then when I am done testing I move it to functions in here which could be imported and used for testing

src/utilities/email-helper.ts
```diff
@@ -44,7 +50,22 @@ class EmailHelper {
...
+                attachments: [
+                    {
+                      filename: 'favicon.png',
+                      path: 'emails/rederly_favicon.png',
+                      cid: 'favicon'
+                    }
+                ]
+            },
+            transport: this.client,
+            send: true,
```
red-tt-bear, line 67: What is this flag for?
Should it be sending whether or not email is disabled?

red-tt-bear: I see, the original logic still stands, still what is the purpose of this flag?

red-tt-bear (approved):

src/utilities/email-helper.ts
PanopticaRising: Good question; this flag is to force sending when in development mode. If it were false, then only the local preview would be generated. This shouldn't matter since our configuration for emails are handled at the start of the function.

PR #278, opened 2020-11-24
Cleanup

red-tt-bear:


PanopticaRising (approved):

package.json
```diff
@@ -36,6 +33,7 @@
     "pug": "^3.0.0",
     "qs": "^6.9.4",
     "sequelize": "^5.22.3",
+    "sequelize-cli": "^5.5.1",
```
red-tt-bear, line 36: Moved this to a normal dependency so it can be run on the prebuilt post pruned code

package.json
```diff
@@ -5,9 +5,6 @@
   "main": "index.js",
   "dependencies": {
     "@hapi/joi": "^15.1.1",
-    "@types/email-templates": "^7.1.0",
```
red-tt-bear, line 8: Moved these three to dev dependencies

Types should not be part of the build
Also acorn is a peer dependency of eslint which is a dev dependency (fixed the versino too)

PR #287, opened 2020-12-03
RED-317 Added warning logs for long running requests

red-tt-bear:


src/server.ts
```diff
@@ -34,15 +35,35 @@ const {
 
 const app = express();
 
-if (configurations.server.logAccess) {
-    app.use(morgan('dev', {
-        stream: {
-            write: (message): void => {
-                logger.info(message);
+app.use(morgan((tokens, req, res) => {
+    const responseTime = parseInt(tokens['response-time'](req, res) ?? '', 10);
```
PanopticaRising, line 39: I don't understand what's happening here. Is `tokens['response-time']` a function that takes the request and response objects?

red-tt-bear: This is pulled directly from their documentation under: `Using a custom format function`
http://expressjs.com/en/resources/middleware/morgan.html

tokens is an object of methods used to fetch details of the request
response-time is hyphenated so it need to be read from bracket notation
It can be null so I default to empty string which will resolve to NaN (which is checked for in the warning)

doombot-exe (approved):

PR #293, opened 2020-12-08
Added postbuild step to copy templates over to the built directory.

PanopticaRising:
We should consider adding `rimraf` instead of `rm -f`


package.json
```diff
@@ -11,8 +11,9 @@
     "boom": "^7.3.0",
     "cls-hooked": "^4.2.2",
     "cookie-parser": "^1.4.5",
+    "cpy-cli": "^3.1.1",
```
red-tt-bear, line 14: This should be a dev dependency

red-tt-bear (approved):

doombot-exe (approved):

PR #305, opened 2020-12-16
Feature/red 324 openlab integration

doombot-exe:
will need to do testing on this with the openlab -- requested to wait until after this week due to finals.


src/features/courses/course-controller.ts
```diff
@@ -3916,6 +3916,58 @@ You should be able to reply to the student's email address (${options.student.em
...
+        // support for TAs receiving notifications
+        const instructor = await course.getInstructor();
+        const instructorEmail = `Prof. ${instructor.lastName} <${instructor.email}>`;
+
+        const problem = question.problemNumber;
+        const problemSetId = topic.name;
+        const courseId = course.name;
+        const problemPath = question.webworkQuestionPath;
+        const email = [instructorEmail];
+        const studentName = user.firstName+' '+user.lastName;
```
red-tt-bear, line 3944: Spacing is a little odd here

src/features/courses/course-controller.ts
```diff
@@ -3916,6 +3916,58 @@ You should be able to reply to the student's email address (${options.student.em
...
+        const emailURL = `${baseURL}/common/courses/${course.id}/topic/${topic.id}/grading?problemId=${question.id}&userId=${user.id}`;
+        const renderResponse = await rendererHelper.getProblem({
+            sourceFilePath: problemPath,
+            problemSeed: grade?.randomSeed,
+            outputformat: OutputFormat.STATIC,
+            showHints: false,
+            showSolutions: false,
+            permissionLevel: 0,
+        });
+        const rawHTML = rendererHelper.cleanRendererResponseForTheResponse(renderResponse as RendererResponse).renderedHTML;
```
red-tt-bear, line 3954: You don't need to clean it, that just deletes the other fields from the object

src/features/courses/course-controller.ts
```diff
@@ -3916,6 +3916,58 @@ You should be able to reply to the student's email address (${options.student.em
...
+        const problem = question.problemNumber;
+        const problemSetId = topic.name;
+        const courseId = course.name;
+        const problemPath = question.webworkQuestionPath;
+        const email = [instructorEmail];
+        const studentName = user.firstName+' '+user.lastName;
+        const emailURL = `${baseURL}/common/courses/${course.id}/topic/${topic.id}/grading?problemId=${question.id}&userId=${user.id}`;
+        const renderResponse = await rendererHelper.getProblem({
+            sourceFilePath: problemPath,
+            problemSeed: grade?.randomSeed,
```
red-tt-bear, line 3948: We should nil check grade so we don't need the question mark here

src/utilities/form-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+class FormHelper {
+    /**
+     * This function converts objects into FormData
+     * @param object the object to be converted
+     * @param unmerge *optional* the subroutine tasked with handling arrays inside `object`
+     * @returns the FormData converted from `object`
+     */
+    objectToFormData = ({
+        object,
+        unmerge,
```
red-tt-bear, line 42: I think we should make unmerge a required property
At very least it should have a default value
```TypeScript
unmerge = unmergeStrategies.unmergeDuplicatingKey
```

I guess I can see it being optional if you think your object won't have any arrays 🤔

src/utilities/openlab-helper.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as Joi from '@hapi/joi';
+import 'joi-extract-type';
+import { isAxiosError } from './axios-helper';
+import logger from './logger';
+import WrappedError from '../exceptions/wrapped-error';
+import { OpenLabRedirectInfo } from '../features/courses/course-types';
+import * as qs from 'qs';
+import formHelper, { unmergeStrategies } from './form-helper';
+
+const openLabAxios = axios.create({
```
red-tt-bear, line 13: I thought we talked about how this was a mistake on the renderer helper and should be in the class that way we could have multiple instances of the helper but use a default singleton?

src/utilities/openlab-helper.ts
```diff
@@ -0,0 +1,85 @@
...
+    baseURL: configurations.openlab.url,
+    responseType: 'text',
+    timeout: configurations.openlab.requestTimeout,
+    params: {
+        webwork: 1,
+    }
+});
+
+const OPENLAB_QS = qs.stringify({webwork: 1});
```
red-tt-bear, line 22: you have this in the instance definition and here

src/utilities/openlab-helper.ts
```diff
@@ -0,0 +1,85 @@
...
+        rawHTML,
+    }: OpenLabRedirectInfo): Promise<OpenLabResponse> {
+        let pgObject = '';
+        if (!_.isNil(rawHTML)) {
+            pgObject = Buffer.from(rawHTML).toString('base64');
+        } else {
+            // die -- we must have the renderedHTML for OpenLab
+            // and we don't have any anonymous ways of id-ing the users involved...
+            logger.error(`OpenLab query requested for ${problemPath} by a student in ${courseId}`);
+            throw new WrappedError('Request to OpenLab came back empty');
```
red-tt-bear, line 52: Should be RederlyError

src/utilities/renderer-helper.ts
```diff
@@ -4,21 +4,20 @@ import Role from '../features/permissions/roles';
...
 import urljoin = require('url-join');
+import formHelper, { unmergeStrategies } from './form-helper';
 
 const rendererAxios = axios.create({
     baseURL: configurations.renderer.url,
     responseType: 'json',
     timeout: configurations.renderer.requestTimeout
 });
 
-// TODO switch over to new endpoint
```
red-tt-bear, line 21: Why did you remove this TODO, it still needs to be done 🤣 
/rendered needs to become /render-api
haven't had time to debug why it doesn't work correctly

red-tt-bear (changes requested):
Various little touches

src/utilities/form-helper.ts
doombot-exe: exactly - there are several instances in the renderer helper where we have a clear grasp on what form-data we're sending and it doesn't require a merge strategy.

src/configurations.ts
```diff
@@ -151,6 +151,10 @@ const configurations = {
         url: readStringValue('RENDERER_URL', 'http://localhost:3000'),
         requestTimeout: readIntValue('RENDERER_REQUEST_TIMEOUT', 75000),
     },
+    openlab: {
+        url: readStringValue('OPENLAB_URL', 'https://openlab.citytech.cuny.edu/ol-webwork/'),
```
red-tt-bear, line 155: we probably don't want to publicly expose the openlab url, this one might be better to leave empty string

The only other one we default is jira which i'm less worried about since it is jira cloud and I like the example

src/features/courses/course-route.ts
```diff
@@ -1041,7 +1065,16 @@ router.get('/:id',
         try {
             const params = req.params as GetCourseRequest.params;
             const course = await courseController.getCourseById(params.id);
-            next(httpResponse.Ok('Fetched successfully', course));
+            const university = await course.getUniversity({
+                where: {
+                    active: true,
+                }
+            });
+            const canAskForHelp = university?.universityName === 'CityTech' ?? false;
```
red-tt-bear, line 1073: This is great for now, though I wonder if there is some unique identifier that belongs to the university rather than our application that we can use

Once we start showing the university name in the application we might update it and then this will break

red-tt-bear: (There's nothing to do here)

red-tt-bear (approved):

src/configurations.ts
doombot-exe: stripped

src/features/courses/course-route.ts
```diff
@@ -642,6 +643,29 @@ router.get('/question/:id/grade',
...
+    // This is a typescript workaround since it tries to use the type extractMap
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, GetQuestionGradeRequest.body, unknown>, res: Response, next: NextFunction) => {
+        if (_.isNil(req.session)) {
+            throw new RederlyError(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+
+        const user = await req.session.getUser();
+        const { id: questionId } = req.params as GetQuestionRequest.params;
+        const baseURL = req.headers['rederly-origin'] as string | undefined; // need this because it incorrectly thinks it can be an array
```
PanopticaRising, line 658: We should probably use an enum or string constants instead of hardcoded strings.

src/features/courses/course-route.ts
PanopticaRising: It would have to be another field in the db, rather than a natural identifier, for this to be a truly OSS solution.

src/utilities/openlab-helper.ts
```diff
@@ -0,0 +1,83 @@
...
+        studentName,
+        emailURL,
+        rawHTML,
+    }: OpenLabRedirectInfo): Promise<OpenLabResponse> {
+        let pgObject = '';
+        if (!_.isNil(rawHTML)) {
+            pgObject = Buffer.from(rawHTML).toString('base64');
+        } else {
+            // die -- we must have the renderedHTML for OpenLab
+            // and we don't have any anonymous ways of id-ing the users involved...
```
PanopticaRising, line 48: We could pass in the student's id from session.

PanopticaRising: Unless logging meta already does that for us.

PanopticaRising (approved):

src/utilities/openlab-helper.ts
red-tt-bear: It does, I think this is perfect

PR #307, opened 2020-12-18
Feature/red 10 tarball import

red-tt-bear:


src/playground/playground-renderer-functions.ts
```diff
@@ -1,5 +1,33 @@
 import logger from '../utilities/logger';
 import rendererHelper from '../utilities/renderer-helper';
+import * as nodePath from 'path';
+
+const isPathAccessibleToRenderer = async () => {
```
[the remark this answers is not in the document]
red-tt-bear: It's a playground file, can fix in the future

src/utilities/string-helper.ts
```diff
@@ -0,0 +1,7 @@
+export const getAllMatches = (pattern: RegExp, str: string): Array<RegExpExecArray> => {
```
PanopticaRising, line 1: Does this have a benefit over https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll ?

doombot-exe (approved):

src/utilities/string-helper.ts
red-tt-bear: Nope, I've just had this problem before matchAll existed and didn't realize there was a better way 🤣 
I'll fix this in another pr, thanks for the heads up!

PR #311, opened 2020-12-21
Feature/use aws v2 while v3 broken

PanopticaRising:


src/features/users/user-controller.ts
```diff
@@ -379,6 +379,7 @@ class UserController {
             });
             emailSent = configurations.email.enabled;
         } catch (e) {
+            // TODO: Does it make sense for this to be rethrown?
```
red-tt-bear, line 382: Yeah I thought about that too,
does it make sense for registration to fail if the email didn't go out

registration itself passed so I didn't rethrow the error

src/utilities/email-helper.ts
```diff
@@ -3,12 +3,10 @@ import logger from '../utilities/logger';
 import nodemailer = require('nodemailer');
 import Mail = require('nodemailer/lib/mailer');
 import _ = require('lodash');
-// There is no type def for sendgrid-transport
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const sgTransport = require('nodemailer-sendgrid-transport');
```
red-tt-bear, line 8: We should make sure to delete npm packages we are no longer using

red-tt-bear (approved):

doombot-exe (approved):

doombot-exe (approved):

PR #316, opened 2020-12-24
Updates to build process

red-tt-bear:
Added db sync configuration

Removed no op script (just moved it into the cli)
Updated github build validation to use packaged code
Updated to use cli noop

Added build validation when generating the build

Upgraded node in the github action from 12 to 14
Upgraded node in the docker file from 10 to 14


PanopticaRising (approved):

doombot-exe (commented):
belated

PR #321, opened 2020-12-29
Feature/red 10 tarball import

red-tt-bear:


src/features/courses/course-controller.ts
```diff
@@ -302,6 +302,10 @@ class CourseController {
             where[`$enrolledStudents.${StudentEnrollment.rawAttributes.userId.field}$`] = options.filter.enrolledUserId;
         }
 
+        if (!_.has(where, 'active')) {
```
PanopticaRising, line 305: If there's no `active` flag, we force only active courses? What about if we want both active and inactive courses?

PanopticaRising (approved):

src/features/courses/course-controller.ts
red-tt-bear: our active flag is soft delete, we should never be fetching inactive anything
technically old courses will remain active

PR #323, opened 2020-12-29
Generic Email Templates.

PanopticaRising:


red-tt-bear (approved):
Looks good but we should fix eslint, I would do in github except that would be a bunch of small commits

PR #341, opened 2021-01-06
Feature/course import auditing

PanopticaRising:


src/database/migrations/20210104113400-course-audit.ts
```diff
@@ -0,0 +1,36 @@
+import { DataTypes } from 'sequelize';
```
red-tt-bear, line 1: Why do we have these as two different imports?
Skeleton will help with this in the future

src/database/migrations/20210104113400-course-audit.ts
```diff
@@ -0,0 +1,36 @@
...
+import { QueryInterface } from 'sequelize';
+
+export default {
+    up: async (queryInterface: QueryInterface): Promise<void> => {
+        // Transactions are automatically use because a namespace is injected into sequelize when fetching configurations
+        await queryInterface.sequelize.transaction(async () => {
+            await queryInterface.addColumn('course_topic_content', 'course_topic_content_errors', {
```
red-tt-bear, line 8: I think it's fine, but adding `count` to the end would make it much more clear what this is doing

src/database/models/course-ww-topic-question.ts
```diff
@@ -20,6 +20,10 @@ export interface CourseWWTopicQuestionInterface {
     courseQuestionAssessmentInfo?: CourseQuestionAssessmentInfo;
 }
 
+export interface CourseTopicQuestionErrors {
```
red-tt-bear, line 23: Nice!

src/features/courses/course-controller.ts
```diff
@@ -729,7 +729,13 @@ class CourseController {
                 }
             });
 
+            // If only one question is deleted, rather than a whole topic.
             if (!_.isNil(existingQuestion)) {
+                // If this question was previously in an error state, decrement the topic's error cache.
+                if (!_.isNil(existingQuestion.errors)) {
+                    existingQuestion.getTopic().then(topic => topic.decrement('errors'));
```
red-tt-bear, line 736: Is decrement defined? I don't see it anywhere

doombot-exe (approved):

src/features/courses/course-controller.ts
red-tt-bear: I see that the build check succeeded, I guess this is a sequelize function?

red-tt-bear: Regardless we should not have orphaned promises without catches
We should add
```TypeScript
.catch(err => logger.error(`Could not decrement topic errors: ${err.message}`));
```

red-tt-bear (approved):

src/features/courses/course-controller.ts
PanopticaRising: It's a sequelize function, yeah. Do we need to explicitly catch errors like those? I thought if they bubbled up, the transaction would be rolled back and we'd still get a Sequelize error on the logs.

PR #344, opened 2021-01-07
Optimized by removing the double-renderer call when doing an archive …

PanopticaRising:
…upload.

### DIFF
--- a/src/features/courses/course-controller.ts
+++ b/src/features/courses/course-controller.ts
@@ -6,7 +6,7 @@ import { BaseError } from 'sequelize';
 import NotFoundError from '../../exceptions/not-found-error';
 import CourseUnitContent from '../../database/models/course-unit-content';
 import CourseTopicContent, { CourseTopicContentInterface } from '../../database/models/course-topic-content';
-import CourseWWTopicQuestion, { CourseWWTopicQuestionInterface } from '../../database/models/course-ww-topic-question';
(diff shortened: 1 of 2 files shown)


doombot-exe (approved):

src/features/courses/course-controller.ts
```diff
@@ -1289,20 +1280,28 @@ class CourseController {
...
+                    let errors: CourseTopicQuestionErrors | null = null;
                     if (pgFilePath === undefined) {
                         throw new NotFoundError('The file path for this problem was not defined.');
                     }
 
-                    // TODO: This call is duplicated for Tarball uploads. Skip it and get the error objects from that call.
-                    if (await rendererHelper.isPathAccessibleToRenderer({problemPath: pgFilePath}) === false) {
-                        errors = {[pgFilePath]: [`${pgFilePath} cannot be found.`]};
-                        courseRepository.incrementTopicErrorCount(options.courseTopicId);
+                    const parsedErrors = (options as CreateQuestionsForTopicFromParsedDefFileOptions).errors;
```
red-tt-bear, line 1292: Why do we need to cast options?

red-tt-bear (approved):

src/features/courses/course-controller.ts
PanopticaRising: Because errors doesn't exist on both objects, and I'm just checking if it exists on line 1293.

PR #347, opened 2021-01-09
Added a CLI command that checks all question/assessmentinfo paths in

PanopticaRising:
active topics and updates that object's associated error object.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -141,7 +141,7 @@ const syncBadPathCounts = async (): Promise<void> => {
             if (count > topic.errors) {
                 logger.debug(`Updating topic ${topic.id} error count from ${topic.errors} to ${count}`);
                 topic.errors = count;
-                topic.save();
+                topic.save().catch(e => console.error(`Failed to save error updates on topic ${topic.id}`, e));
             }
         });
     });
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -90,6 +90,8 @@ const syncBadPathCounts = async (): Promise<void> => {
 
         logger.info(`Checking ${topics.length} topics`);
 
+        const allTopicsProcessedPromises: Promise<CourseTopicContent>[] = [];
+
         await topics.asyncForEach(async (topic) => {
             // For each question in the topic, sum the existence of a path problem.
             let count = 0;
@@ -112,7 +114,7 @@ const syncBadPathCounts = async (): Promise<void> => {
 
                 if (topic.topicTypeId === 2) {
                     if (_.isNil(courseQuestionAssessmentInfo)) {
-                        logger.error(`Topic ${topic.id} is an exam but this question has no assessment info.`);
+                        logger.debug(`Topic ${topic.id} is an exam but this question has no assessment info.`);
                         if (isDefinitelyBad) count += 1;
                         return;
                     }
@@ -141,11 +143,13 @@ const syncBadPathCounts = async (): Promise<void> => {
             if (count > topic.errors) {
                 logger.debug(`Updating topic ${topic.id} error count from ${topic.errors} to ${count}`);
                 topic.errors = count;
-                topic.save().catch(e => console.error(`Failed to save error updates on topic ${topic.id}`, e));
+                allTopicsProcessedPromises.push(topic.save());
             }
         });
+        await Promise.all(allTopicsProcessedPromises);
     });
-    console.info(`Request took ${performance.now() - time}`);
+
+    logger.info(`Request took ${performance.now() - time}`);
 };
 
 const commandLookup: {[key: string]: () => unknown} = {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -16,8 +16,11 @@ if (configurations.email.enabled) {
 import { sync } from './database';
 import courseController from './features/courses/course-controller';
 import StudentWorkbook from './database/models/student-workbook';
-import rendererHelper from './utilities/renderer-helper';
+import rendererHelper, {VALID_PG_PATH_REGEX} from './utilities/renderer-helper';
 import { useDatabaseTransaction } from './utilities/database-helper';
+import CourseTopicContent from './database/models/course-topic-content';
+import CourseWWTopicQuestion from './database/models/course-ww-topic-question';
+import CourseQuestionAssessmentInfo from './database/models/course-question-assessment-info';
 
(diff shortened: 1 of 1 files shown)


src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
     logger.info('noop performed');
 };
 
+
+
+const syncBadPathCounts = async (): Promise<void> => {
+    await useDatabaseTransaction(async () => {
+        const topics = await CourseTopicContent.findAll(
```
red-tt-bear, line 69: I was thinking this would search for bad problem and questions and then fetch their respective topics
But since this should only be run once I think it is fine

src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
...
+                const {webworkQuestionPath, errors, courseQuestionAssessmentInfo} = question;
+                const additionalProblemPaths = courseQuestionAssessmentInfo?.additionalProblemPaths;
+                const additionalErrors = courseQuestionAssessmentInfo?.errors;
+
+                const isDefinitelyBad = VALID_PG_PATH_REGEX.test(webworkQuestionPath) === false;
+
+                // If this is definitely bad and not already in the error object, add it.
+                if (isDefinitelyBad && (_.isNil(errors) || _.isEmpty(errors) || !(webworkQuestionPath in errors))){
+                    logger.debug(`Updating ${question.id} from topic ${topic.id} with a path error.`);
+                    question.errors = {
```
red-tt-bear, line 104: If there was already errors wouldn't this override it
Shouldn't question.errors be spread into the object in case it is not null?

src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
...
+                        [webworkQuestionPath]: [`${webworkQuestionPath} cannot be found.`]
+                    };
+                    question.save();
+                }
+
+                if (topic.topicTypeId === 2) {
```
red-tt-bear, line 110: I don't know if we need this check, if it has assessment info since someone went back and forth we might still want those errors

I don't feel too strongly about this

src/cli.ts
PanopticaRising: Oh, this is a good catch. Right now, it doesn't matter because path errors are the only errors we use this field for, but you're right.

src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
...
+                        if (isDefinitelyBad && (_.isNil(additionalErrors) || _.isEmpty(additionalErrors) || !(path in additionalErrors))) {
+                            if (_.isNil(courseQuestionAssessmentInfo.errors)) {
+                                courseQuestionAssessmentInfo.errors = {};
+                            }
+                            logger.debug(`Updating ${courseQuestionAssessmentInfo.id} from topic ${topic.id} with an additional path error.`);
+                            courseQuestionAssessmentInfo.errors[path] = [`${path} cannot be found`];
+                        }
+                    });
+                    courseQuestionAssessmentInfo.save();
+                    if (isDefinitelyBad || atLeastOneIsBad) count += 1;
```
red-tt-bear, line 130: I thought we were counting the number of questions that had errors, not the number of errors on the question

If I had one question with 3 paths for an exam and they were all bad, shouldn't that only be one error on the topic?

This looks like it would add 3 which would be the number of bad paths, I'm fine either way as long as we are consistent

src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
...
+                    if (isDefinitelyBad) count += 1;
+                }
+            });
+
+            // Count can be equal to or less than because it doesn't include full renderer checks.
+            // Questions will still be annotated correctly with updated error paths.
+            if (count > topic.errors) {
+                logger.debug(`Updating topic ${topic.id} error count from ${topic.errors} to ${count}`);
+                topic.errors = count;
+                topic.save();
```
red-tt-bear, line 141: Not awaiting

red-tt-bear: If we don't await we should have a .catch to avoid orphaned promise errors

src/cli.ts
PanopticaRising: We don't want these errors because they'd be displayed on the details page. We don't want to force people to have to clear their assessment paths if they're not using them.

src/cli.ts
PanopticaRising: This should be outside of the `asyncForEach`, so it should only increment the topic.errors **once** if either the question path or one of the assessment paths are bad.

src/cli.ts
```diff
@@ -59,10 +62,93 @@ const noop = (): void => {
...
+                    logger.debug(`Updating ${question.id} from topic ${topic.id} with a path error.`);
+                    question.errors = {
+                        [webworkQuestionPath]: [`${webworkQuestionPath} cannot be found.`]
+                    };
+                    question.save();
+                }
+
+                if (topic.topicTypeId === 2) {
+                    if (_.isNil(courseQuestionAssessmentInfo)) {
+                        logger.error(`Topic ${topic.id} is an exam but this question has no assessment info.`);
```
PanopticaRising, line 115: @red-tt-bear Should this be an error log? I don't know if we create all the assessmentinfos at once or only as-needed.

src/cli.ts
PanopticaRising: Does that still work for cancelling the transaction?

src/cli.ts
red-tt-bear: I believe it is as needed, so probably debug log

src/cli.ts
red-tt-bear: I don't think it will cancel the transaction period since it is orphaned, it might 1 level deep i'm not sure, why don't we await this?

src/cli.ts
```diff
@@ -59,10 +63,96 @@ const noop = (): void => {
...
+            // Count can be equal to or less than because it doesn't include full renderer checks.
+            // Questions will still be annotated correctly with updated error paths.
+            if (count > topic.errors) {
+                logger.debug(`Updating topic ${topic.id} error count from ${topic.errors} to ${count}`);
+                topic.errors = count;
+                topic.save();
+            }
+        });
+    });
+    console.info(`Request took ${performance.now() - time}`);
```
red-tt-bear, line 148: Should be logger.info
I wonder why we didn't get an eslint warning

red-tt-bear (approved):

PR #361, opened 2021-01-19
Bugfix/red 380 show errors when uploading

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -1385,7 +1385,7 @@ class CourseController {
         let lastProblemNumber = await courseRepository.getLatestProblemNumberForTopic(options.courseTopicId) || 0;
 
 
-        const topic = await CourseTopicContent.findOne({where: {id: options.courseTopicId}});
+        const topic = options.topic ?? await CourseTopicContent.findOne({where: {id: options.courseTopicId}});
```
red-tt-bear, line 1388: Since you can have a mismatch in topic id and topic it might make sense to check and warn
After this line you can just include:
```Typescript
if (topic.id !== options.courseTopicId) logger.warn('Id mismatch')
```
If we were throwing an error it would make sense to nil check and use the raw values before the query was made

red-tt-bear (approved):

doombot-exe (approved):

PR #365, opened 2021-01-20
Added email to verification call.

PanopticaRising:


src/features/users/user-controller.ts
```diff
@@ -436,17 +436,19 @@ class UserController {
...
+        } else if (user.verifyToken !== verifyToken) {
             throw new IllegalArgumentException('Invalid verification token');
-        } else if(moment().isAfter(user.verifyTokenExpiresAt)) {
+        } else if (moment().isAfter(user.verifyTokenExpiresAt)) {
             throw new IllegalArgumentException('Verification token has expired');
-        } else if(user.verified) {
+        } else if (user.verified) {
             logger.warn('Verification token should be set to null on verify, thus this should not be possible');
             throw new IllegalArgumentException('Already verified');
+        } else if (confirmEmail !== user.email) {
```
red-tt-bear, line 450: I personally prefer keeping things like this obscure for security reasons but I'm ok with it

I would have just added this to getUserByVerifyToken so it would return null if the combo was not correct

red-tt-bear (approved):

doombot-exe (approved):

PR #375, opened 2021-01-27
ignore hardcoded URL images

doombot-exe:


PanopticaRising (approved):

src/utilities/webwork-utilities/importer.ts
```diff
@@ -96,9 +96,9 @@ const perlQuotes: Array<[string, string]> = [
 const imageInPGFileRegex = new RegExp(
     [
         '(?<!#.*)(?:', // Comment, using non capture group to spread amongst or
-        `(?:image\\s*\\(\\s*(${perlQuotes.map(perlQuote => `${perlQuote[0]}.+?${perlQuote[1]}`).join('|')})\\s*(?:,(?:\\s|.)*?)?\\))`, // image call
+        `(?:image\\s*\\(\\s*(${perlQuotes.map(perlQuote => `${perlQuote[0]}(?!\\s*https?:).+?${perlQuote[1]}`).join('|')})\\s*(?:,(?:\\s|.)*?)?\\))`, // image call
```
red-tt-bear, line 99: We need to add a starts with if we can in case the image has https in it
Also missing `//`

red-tt-bear: I'm wrong about the starts with
However, I still feel like we should add the //
I could have a file named https:tom.png and that is valid and clearly not a url

red-tt-bear (approved):

src/utilities/webwork-utilities/importer.ts
doombot-exe: it's already 'starts with' (with the only exception being whitespace, which is unnecessary?)

PR #385, opened 2021-01-30
Feature/bulk pdf export

PanopticaRising:


assets/pdf.pug
```diff
@@ -0,0 +1,14 @@
+doctype html
```
PanopticaRising, line 1: This is unused

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "@hapi/joi": "^15.1.1",
+    "@types/html5-to-pdf": "^3.1.1",
```
PanopticaRising, line 8: Unused

src/features/courses/course-route-validation.ts
```diff
@@ -623,3 +623,23 @@ export const getVersionValidation = {
...
+    body: {},
+};
+
+export const endBulkExportValidation = {
+    params: {
+        topicId: Joi.number().required(),
+    },
+    query: {},
+    body: {
+        exportUrl: Joi.string().required(),
```
PanopticaRising, line 643: This needs to be optional.

src/playground/playground.ts
```diff
@@ -3,6 +3,7 @@ import '../extensions';
 // TODO change logger to just use console in this case
 import logger from '../utilities/logger';
 import '../global-error-handlers';
+import courseController from '../features/courses/course-controller';
```
PanopticaRising, line 6: This can be reverted.

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,261 @@
...
+            problemAttachments: any;
+            randomSeed: number;
+        }[];
+    }[];
+}
+
+export default class ExportPDFHelper {
+    bulkExportAxios = axios.create({
+        baseURL: configurations.bulkPdfExport.baseUrl,
+        timeout: 120000
```
PanopticaRising, line 57: This timeout doesn't need to be so high since we're asynchronous now.

src/database/models/course-topic-content.ts
```diff
@@ -32,6 +32,9 @@ export default class CourseTopicContent extends Model implements CourseTopicCont
     public deadDate!: Date;
     public partialExtend!: boolean;
 
+    public lastExported!: Date | null;
```
red-tt-bear, line 35: Why not embed this in the export url?

src/extensions/array-extension.ts
```diff
@@ -12,6 +13,14 @@ Array.prototype.asyncForEach = function <T>(callbackfn: (value: T, index: number
     return Promise.all(promises);
 };
 
+Array.prototype.sequentialAsyncForEach = async function <T>(callbackfn: (value: T, index: number, array: T[]) => Promise<unknown>): Promise<unknown[]> {
```
red-tt-bear, line 16: Great function name

src/features/courses/course-route.ts
```diff
@@ -307,6 +309,78 @@ router.get('/topic/:topicId/version/:userId',
...
+    asyncHandler(async (req: RederlyExpressRequest<any, unknown, EndBulkExportRequest.body, EndBulkExportRequest.query>, _res: Response, next: NextFunction) => {
+        const topic = await CourseTopicContent.findOne({
+            where: {
+                id: req.params.topicId,
+                active: true,
+            },
+        });
+
+        if (_.isNil(topic)) {
+            throw new NotFoundError('A endExport PUT was received for a non-existant topic. Was it deleted?');
```
red-tt-bear, line 326: An*? :trollface:

red-tt-bear: This error would get shown on the frontend, we might want to clean up the message

red-tt-bear: nvm, it's called from the microservice*

src/features/courses/course-route.ts
```diff
@@ -307,6 +309,78 @@ router.get('/topic/:topicId/version/:userId',
...
+        }
+
+        if (_.isNil(req.body.exportUrl)) {
```
red-tt-bear, line 329: Why would this be nil?

src/features/courses/course-route.ts
```diff
@@ -307,6 +309,78 @@ router.get('/topic/:topicId/version/:userId',
...
+            topic.exportUrl = null;
+            topic.lastExported = null;
+        } else {
+            topic.exportUrl = req.body.exportUrl;
+        }
+        await topic.save();
+
+        next(httpResponse.Ok('Got it!'));
```
red-tt-bear, line 337: I like the message! Feels like an inspirational poster

src/features/courses/course-route.ts
```diff
@@ -307,6 +309,78 @@ router.get('/topic/:topicId/version/:userId',
...
+            throw new NotFoundError('No such topic was found.');
+        }
+
+        // Since this is a job, we'll always have this by here.
+        const exportDetails = {lastExported: topic.lastExported, exportUrl: topic.exportUrl};
+
+        const helper = new ExportPDFHelper();
+
+        // If we're just checking, return what's already here.
+        if (query.force === false) {
```
red-tt-bear, line 372: NIT: This feels like it should be multiple routes, maybe with a common middleware

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            problemAttachments: any;
+            randomSeed: number;
+        }[];
+    }[];
+}
+
+export default class ExportPDFHelper {
+    bulkExportAxios = axios.create({
+        baseURL: configurations.bulkPdfExport.baseUrl,
```
red-tt-bear, line 58: NIT: in the future I want these configurations to be parameters for the constructor
We would then use the configurations for a default instance

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+    });
+
+    shouldStartNewExport = (topic: CourseTopicContent): boolean => {
+        // If the time is within 5 minutes, do not allow a new export
+        if (!_.isNil(topic.lastExported) && moment().isBefore(moment(topic.lastExported).add(5, 'minutes'))) {
```
red-tt-bear, line 63: Do we let the user know of this 5-minute restriction?

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+            return false;
+        }
+
+        return true;
+    }
+    
+    start = async ({topic, professorUUID}: {topic: CourseTopicContent; professorUUID: string}): Promise<void> => {
+        topic.lastExported = new Date();
+        topic.exportUrl = null;
+        await topic.save();
```
red-tt-bear, line 73: I prefer for these microservice helpers to not make calls the to the database

Reason being is that I think of them as sdks

If we ever wanted to use this in another application or external script we wouldn't be able to

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+                lastName: user.lastName,
+                problems: studentGrades,
+                professorUUID: professorUUID
+            };
+            return await this.bulkExportAxios.post('export/', postObject);
+        });
+        
+        // Request zip
+        try {
+            return await this.bulkExportAxios.get(`export/?profUUID=${professorUUID}&topicId=${topic.id}`);
```
red-tt-bear, line 235: Why are you manually doing the query parameters and not using the params option for axios?

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+        } catch (e) {
+            console.error(e);
```
red-tt-bear, line 237: logger

src/database/models/course-topic-content.ts
red-tt-bear: I guess because the export url doesn't exist until the export is done

src/extensions/object-extension.ts
```diff
@@ -0,0 +1,16 @@
+// interface Object {
```
red-tt-bear, line 1: I guess you gave up on making this an extension 🤣

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+    };
+    professorUUID: string;
+    problems: {
+        number: number;
+        srcdoc: string;
+    }[];
+}
+
+// topic from db
+type FROM_DB = {
```
red-tt-bear, line 34: Why is this all caps?

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+
+                const obj = {
+                    sourceFilePath: gradeInstanceAttachments?.studentGradeInstance?.webworkQuestionPath ?? question.webworkQuestionPath,
+                    problemSeed: gradeInstanceAttachments?.randomSeed ?? gradeInstanceAttachments?.studentGradeInstance?.randomSeed ?? grade.randomSeed,
+                    formData: gradeInstanceAttachments?.submitted.form_data,
+                    showCorrectAnswers: true,
+                    answersSubmitted: 1,
+                    outputformat: rendererHelper.getOutputFormatForRole(Role.PROFESSOR),
+                    permissionLevel: rendererHelper.getPermissionForRole(Role.PROFESSOR),
+                    showSolutions: true,
```
red-tt-bear, line 181: Does show solutions work since they are collapsible?

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+                };
+
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
red-tt-bear, line 184: Should have a comment as to why we are disabling the rule

red-tt-bear (approved):

src/database/models/course-topic-content.ts
PanopticaRising: We needed something to differentiate when the process to export has never started and when it is in progress.

src/features/courses/course-route.ts
PanopticaRising: In the case of an error in the microservice.

src/utilities/export-pdf-helper.ts
PanopticaRising: I don't have a notification in for that yet. I was thinking about something like 'Please wait a few minutes before re-creating the export'. We may want to just drop the limit with the verbiage change.

src/utilities/export-pdf-helper.ts
```diff
@@ -0,0 +1,242 @@
...
+                                        active: true,
+                                    }
+                                }
+                            ]
+                        },
+                    ]
+                });
+
+                const obj = {
+                    sourceFilePath: gradeInstanceAttachments?.studentGradeInstance?.webworkQuestionPath ?? question.webworkQuestionPath,
```
doombot-exe, line 174: With the problem path having been added to the workbook - does it make sense here to rely on the workbook as the source of truth? Then falling back to grade instance, and finally to the topic question as a last resort?

src/utilities/export-pdf-helper.ts
doombot-exe: Hints and solutions are obscured as 'knowls'
They are wrapped in an <a> tag, with class 'internal' 
The addition of 'active' to the class causes the hint/solution to be rendered on the page.

doombot-exe (approved):

src/features/courses/course-route.ts
red-tt-bear: You might wanna add a comment for that

PR #408, opened 2021-02-16
RED-485 Fixed reply to from email students

red-tt-bear:
Used email instead of preferred email
Also added a poor man's template to give instructions to the student

### DIFF
--- a/src/features/users/user-controller.ts
+++ b/src/features/users/user-controller.ts
@@ -169,13 +169,22 @@ class UserController {
         // I tried bcc: users.map(user => user.email) but I got an error that an email address was required
         const emailPromises = [];
         for (let i = 0; i < users.length; i++) {
+            const user = users[i];
+            const poorMansTemplate = `
+            Hello ${user.firstName} ${user.lastName},
+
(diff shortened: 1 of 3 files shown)


src/features/users/user-controller.ts
```diff
@@ -169,13 +169,22 @@ class UserController {
         // I tried bcc: users.map(user => user.email) but I got an error that an email address was required
         const emailPromises = [];
         for (let i = 0; i < users.length; i++) {
+            const user = users[i];
+            const poorMansTemplate = `
```
PanopticaRising, line 173: :trollface:

src/features/users/user-controller.ts
```diff
@@ -169,13 +169,22 @@ class UserController {
...
+            Hello ${user.firstName} ${user.lastName},
+
+            You have received a message through Rederly:
```
PanopticaRising, line 176: I'm iffy on the `through Rederly` part. It feels misleading because we don't have an in-app messaging system. But I don't have a better suggestion right now.

PanopticaRising (approved):

PR #411, opened 2021-02-18
Feature/expand single print options

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -4505,24 +4543,33 @@ You can contact your student at ${options.student.email} or by replying to this
                             as: 'studentGradeInstance',
                             attributes: ['id', 'webworkQuestionPath'],
                             where: {
-                                active: true,
+                                active: false,
```
red-tt-bear, line 4546: Why do we want active false?

red-tt-bear (changes requested):
active false

src/features/courses/course-controller.ts
PanopticaRising: Good catch, this was meant to be `required: false`.

src/features/courses/course-controller.ts
```diff
@@ -4512,17 +4550,27 @@ You can contact your student at ${options.student.email} or by replying to this
...
                                     }
                                 }
                             ]
                         },
                     ]
-                });
+                }) : null;
+                
+                let attachments = gradeInstanceAttachments?.studentGradeInstance?.problemAttachments;                
+                if (mainData.topicTypeId === 1) {
```
red-tt-bear, line 4564: I'm surprised you don't use a ternary here

red-tt-bear (approved):

PR #431, opened 2021-03-11
Added status to the backend

red-tt-bear:


src/middleware/status-handler.ts
```diff
@@ -0,0 +1,163 @@
...
+            alreadyCrawling = true;
+            statusPromises = statusAccessibleOptions.map(statusAccessibleOption => checkAccessible(statusAccessibleOption));
+            healthPromises = healthAccessibleOptions.map(healthAccessibleOption => checkAccessible(healthAccessibleOption));
+        }
+        const customPromises = customChecks.map(customCheck => customCheck.call());
+
+        const statusResults = await Promise.all(statusPromises);
+        const healthResults = await Promise.all(healthPromises);
+        const customResults = await Promise.all(customPromises);
+        const allResults = [...statusResults, ...healthResults, ...customResults];
```
PanopticaRising, line 125: Should these be `allSettled`, or will checkAccessible always handle it?

PanopticaRising (approved):

src/middleware/status-handler.ts
red-tt-bear: checkAccessible has a catch and thus cannot reject

PR #435, opened 2021-03-12
Feature/red 425 hiding exam results gib

PanopticaRising:


src/features/courses/course-types.ts
```diff
@@ -281,6 +281,7 @@ export interface GetStatisticsOnUnitsOptions {
     where: {
         courseId?: number;
         userId?: number;
+        userRole: Role;
```
PanopticaRising, line 284: This was an artifact and can probably be removed.

src/features/courses/course-controller.ts
```diff
@@ -3009,6 +3023,13 @@ class CourseController {
             [`$questions.grades.${StudentGrade.rawAttributes.userId.field}$`]: userId,
         }).omitBy(_.isNil).value() as sequelize.WhereOptions;
 
+        if (userRole === Role.STUDENT) {
+            // TODO: Fix this.
```
PanopticaRising, line 3027: The typing on this was a little weird. Above, in the other call, it was explicitly set to WhereOptions but it was inferred as WhereAttributesHash so I explicitly tagged this as one.

red-tt-bear (approved):

PR #436, opened 2021-03-15
feat: Added weights per problemt to be passed to the frontend to disp…

PanopticaRising:
…lay to users.

feat: Added Assessment info to be passed back to devlier better error messages.


src/features/courses/course-controller.ts
```diff
@@ -101,6 +101,14 @@ class CourseController {
                                 ...(userId && {userId: userId}),
                             },
                             required: false,
+                        },
+                        {
+                            model: TopicAssessmentInfo,
```
red-tt-bear, line 106: What are we using this for?

PR #437, opened 2021-03-19
Feature/add description to topics

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -5151,7 +5151,19 @@ You can contact your student at ${options.student.email} or by replying to this
             return row;
         });
     }
-
+    
+    async addFeedback({
+        workbookId,
+        content
+    }: AddFeedbackOptions): Promise<void> {
+        StudentWorkbook.update({
```
red-tt-bear, line 5159: Not returning so can't do anything with the result
More importantly not awaiting so it will always succeed

src/features/courses/course-route-validation.ts
```diff
@@ -658,3 +661,14 @@ export const endBulkExportValidation = {
         exportUrl: Joi.string().optional(),
     },
 };
+
+export const postFeedbackValidation = {
+    params: {
+    },
+    query: {
+        workbookId: Joi.number().required(),
```
red-tt-bear, line 669: this should probably be in the path 🤷‍♂️

src/features/courses/course-route.ts
```diff
@@ -1640,4 +1640,18 @@ router.post('/question/editor/catalog',
             problems: Object.keys(result).filter(elm => elm.endsWith('.pg'))
         }));
     }));
+
+router.post('/feedback', 
+    authenticationMiddleware,
+    validate(postFeedbackValidation),
+    asyncHandler(async (req: RederlyExpressRequest<PostFeedbackRequest.params, unknown, PostFeedbackRequest.body, unknown>, _res: Response, next: NextFunction) => {
+        const res = courseController.addFeedback({
```
red-tt-bear, line 1648: your not awaiting here, so no error handling and the response is probably garbage

red-tt-bear (changes requested):
2 awaits

red-tt-bear (approved):

PR #439, opened 2021-03-31
Feature/red 443 blocking grades

PanopticaRising:


src/features/courses/course-controller.ts
```diff
@@ -2857,33 +2871,121 @@ class CourseController {
...
+                    ]
+                }
+            ]
+        });
+
+        let attributes;
+        let gradeInstanceAttributes;
+        let workbookAttributes;
+        if (userRole === Role.STUDENT && (topicAssessmentInfo?.showTotalGradeImmediately === false || topicAssessmentInfo?.showItemizedResults === false)) {
+            logger.warn('Setting attributes to empty');
```
PanopticaRising, line 2918: Gotta remove this.

src/features/courses/course-controller.ts
```diff
@@ -2857,33 +2871,130 @@ class CourseController {
...
+                        }
+                    ]
+                }
+            ]
+        });
+
+        let attributes;
+        let gradeInstanceAttributes;
+        let workbookAttributes;
+        if (userRole === Role.STUDENT && (topicAssessmentInfo?.showTotalGradeImmediately === false || topicAssessmentInfo?.showItemizedResults === false)) {
```
red-tt-bear, line 2917: Shouldn't this be

userRole is student AND
showTotalGradeImmediately AND
showItemizedResults

The or with the falses seems off

red-tt-bear: or maybe it should just be that they are true? unless there is a double negative somewhere?

red-tt-bear: Oh unless you're using this to filter the attributes and relying on them being undefined if not entered here to show everything?

Without types on the variables it's pretty confusing

red-tt-bear (approved):

PR #448, opened 2021-04-08
Feature/red 545 mem check

red-tt-bear:


src/configurations.ts
```diff
@@ -122,6 +122,14 @@ const configurations = {
         configSalt: readStringValue('CONFIG_SALT', ''),
         autoDeleteTemp: readBooleanValue('AUTO_DELETE_TEMP_FILES', true),
     },
+    monitoring: {
+        memory: {
```
PanopticaRising, line 126: Is there a way to disable memory monitoring?

src/daemons.ts
```diff
@@ -0,0 +1,86 @@
+import logger from './utilities/logger';
+import { getHeapStatistics } from 'v8';
+import configurations from './configurations';
+const {
+    debugThreshold,
+    warningThreshold,
+    errorThreshold,
+    interval: memoryDaemonInterval
+} = configurations.monitoring.memory;
+const minThreshold = Math.min(debugThreshold, warningThreshold, errorThreshold);
```
PanopticaRising, line 10: Is the expected way to avoid log output to set all the environment variables to 100?

red-tt-bear: Yup

src/configurations.ts
red-tt-bear: Set all levels above 100

PanopticaRising (approved):

PR #464, opened 2021-05-04
Stage

red-tt-bear:


src/features/courses/course-route.ts
```diff
@@ -254,6 +255,17 @@ router.get('/:courseId/topic-grades',
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     asyncHandler(async (req: RederlyExpressRequest<any, unknown, GetGradesForTopicsByCourseRequest.body, GetGradesForTopicsByCourseRequest.query>, res: Response, next: NextFunction) => {
         const params = req.params as GetGradesForTopicsByCourseRequest.params;
+
+        if (_.isNil(req.session)) {
+            throw new RederlyError(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+        
+        const user = await req.session.getUser();
```
red-tt-bear, line 263: This logic is wrong
it should be:
```
req.rederlyUserRole === Role.STUDENT
```

That way impersonation works as expected

it also avoids an extra db call, we already fetched the user as well

PanopticaRising (approved):

PR #467, opened 2021-05-06
Feature/red 555 optimize retro

red-tt-bear:


src/database/migrations/20210504010450-retro-updates.ts
```diff
@@ -0,0 +1,46 @@
+// TODO remove from actual migrations, DataTypes will be used
```
PanopticaRising, line 1: Forgot to delete!

src/database/migrations/20210504010450-retro-updates.ts
```diff
@@ -0,0 +1,46 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { QueryInterface, DataTypes } from 'sequelize';
+
+export default {
+  up: async (queryInterface: QueryInterface): Promise<void> => {
+    // Transactions are automatically use because a namespace is injected into sequelize when fetching configurations
+    await queryInterface.sequelize.transaction(async () => {
+      await queryInterface.addColumn('course_topic_content', 'course_topic_content_grade_ids_that_need_retro', {
```
PanopticaRising, line 9: Woof, really trying to hit that 64 character limit. I can see this causing problems for sequelize.

src/extensions/lodash-mixins.ts
```diff
@@ -66,5 +70,54 @@ _.mixin({
...
+        d: 5
+    });
+    ```
+    Expected result
+    ```
+    a 1
+    b 4
+    c 3
+    d 5
+    ```
```
PanopticaRising, line 106: Isn't this the same behavior as assign?
```
Object.assign({
        a: 1,
        b: 2,
        c: 3,
    }, {
        a: 1,
        b: 4,
        d: 5
    })
Object { a: 1, b: 4, c: 3, d: 5 }
```

src/features/courses/course-route.ts
```diff
@@ -445,7 +444,38 @@ router.put('/topic/extend',
             next(httpResponse.Ok('Extended topic successfully', updatesResult));
         }));
 
-router.put('/topic/:id',
+router.put('/topic/:id/regrade',
+    authenticationMiddleware,
+    validate(regradeCourseTopicValidation),
+    paidMiddleware('Regrading topic'),
+    // This is due to a typescript issue where the type mismatches extractMap
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
PanopticaRising, line 452: Unnecessary log

src/features/courses/course-controller.ts
```diff
@@ -202,6 +208,21 @@ class CourseController {
...
             }
 
+            if (includeGradeIdsThatNeedRegrade) {
+                questionSubInclude.push({
+                    model: StudentGrade,
+                    as: 'grades',
+                    required: false,
+                    attributes: ['id'],
+                    where: {
+                        [sequelize.Op.and]: [
```
PanopticaRising, line 218: Is this `and` necessary? It doesn't seem to be doing anything.

src/features/courses/course-controller.ts
```diff
@@ -1902,41 +1996,44 @@ class CourseController {
         gradeResult,
         submitted,
         timeOfSubmission,
-        problemPath
+        problemPath,
+        saveGrade = true,
     }: SetGradeFromSubmissionOptions): Promise<StudentWorkbook | undefined> => {
         return useDatabaseTransaction(async (): Promise<StudentWorkbook | undefined> => {
+            const studentGradeData = _.cloneDeep(studentGrade.get()) as StudentGradeInterface;
```
PanopticaRising, line 2003: Should this be a `get plain` instead of deep cloning? It'll still have all the sequelize methods with the deep clone, won't it?

PanopticaRising (approved):

src/database/migrations/20210504010450-retro-updates.ts
red-tt-bear: 🤣 got me

src/database/migrations/20210504010450-retro-updates.ts
red-tt-bear: Yeah, I was worried about that as well, luckily it will only be a problem if we do a tons of joins and we really need to get rid of the prefix asap

src/database/migrations/20210504010450-retro-updates.ts
```diff
@@ -0,0 +1,46 @@
...
+        type: DataTypes.ARRAY(DataTypes.INTEGER),
+        allowNull: false,
+        defaultValue: [],
+      });
+
+      await queryInterface.addColumn('course_topic_content', 'course_topic_content_retro_started_time', {
+        field: 'course_topic_content_retro_started_time',
```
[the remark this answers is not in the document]
red-tt-bear: It is but it doesn't do anything, I can remove that

src/extensions/lodash-mixins.ts
red-tt-bear: Nope, I talked about this in slack  bit

If I do the assign you have listed it will do
```Typescript
func (param1, param2)
param1.a = param2.a;
param1.b = param2.b;
param.d = param2.d;
```

However this function does
```Typescript
func (param1, param2)
param1.b = param2.b;
param.d = param2.d;
```
Notice that `param1.a = param2.a` is omitted because it is unchanged


This is extremely important because modifying a sequelize result is actually just an interface, the data is sitting in the underlying `dataValues` property
So if you do: `myModel.a = myModel.a`
Then `myModel.changes()` results in `['a']`

We can optimize calls by using these assign changes and only calling save if (`myModel.changes()`)

src/features/courses/course-route.ts
red-tt-bear: Not unnecessary, if we need to break a linting rule we should disable it for the line and explain it

src/features/courses/course-controller.ts
red-tt-bear: It is necessary because there are an array of conditions for the same property

Also this is wrong and should be an or

src/features/courses/course-controller.ts
red-tt-bear: I used `plain: true` because I had seen that in answers before, however `plain: true` is actually default
https://sequelize.org/master/class/lib/model.js~Model.html
> When true, the first returned value of aggregateFunction is cast to dataType and returned. If additional attributes are specified, along with group clauses, set plain to false to return all values of all returned rows. Defaults to true

As mentioned in an above comment sequelize model results are actually interfaces with observers for each of the properties where the values are stored in dataValues, however typescript doesn't expose dataValues so you have to call `.get()`

I thought `.get()` was returning an object with sequelize bindings stripped but it's actually just returning the subobject dataValues, so if you change a value in .get it changes in the parent interface however since we didn't go through the interface `.changed()` does not update

That's why I'm deep cloning it, so the actual model is remaining unchanged

src/features/courses/course-route.ts
red-tt-bear: Oh I misread, yeah I'll remove this

PR #474, opened 2021-05-13
Feature/lti add authentication

PanopticaRising:


tommy-lettieri (approved):

src/server.ts
```diff
@@ -155,6 +157,9 @@ app.use((req, res, next) => {
     next();
 });
 
+
+app.use(basePath + '/lti', lti.app);
```
tommy-lettieri, line 161: this is dangerous
I'm not sure but I feel like if base path ends in `/` this might break

PR #479, opened 2021-05-15
feat: Added startTime and submissionTime to send to the exporter.

PanopticaRising:


src/utilities/export-pdf-helper.ts
```diff
@@ -200,7 +202,9 @@ export default class ExportPDFHelper {
                     })),
                     effectiveScore: grade.effectiveScore,
                     legalScore: grade.legalScore,
-                    srcdoc: src?.renderedHTML ?? 'Could not render this problem.'
+                    srcdoc: src?.renderedHTML ?? 'Could not render this problem.',
+                    startTime: gradeInstanceAttachments?.studentGradeInstance?.createdAt,
+                    submissionTime: gradeInstanceAttachments?.time,
```
red-tt-bear, line 207: This seems very wrong

This should be based on the workbook not the attachments

red-tt-bear: Oh the variable is just badly named
```Typescript
const gradeInstanceAttachments = await StudentWorkbook.findOne({
```

I was looking at the real attachments variable
```Typescript
let attachments = gradeInstanceAttachments?.studentGradeInstance?.problemAttachments;
```

red-tt-bear (approved):

PR #493, opened 2021-05-25
Feature/red 607 less grades

red-tt-bear:


src/features/courses/course-controller.ts
```diff
@@ -3130,26 +3138,26 @@ class CourseController {
...
+            const fetchedEnrollment = await StudentEnrollment.findOne({
+                where: {
+                    courseId: options.courseId,
+                    userId: options.userId,
+                }
+            });
+
+            if (!_.isNil(fetchedEnrollment)) {
+                if (!_.isNil(fetchedEnrollment.dropDate)) {
+                    if (options.reEnrollIfDropped ?? false) {
```
PanopticaRising, line 3151: Is the `?? false` needed? Seems weird to not at least be `=== true`.

PanopticaRising, line 3151: nit: Could probably condense the first two into one check, and convert the `if x else throw` into a `if !x throw`.

PanopticaRising (approved):

PR #494, opened 2021-05-29
RED-611 Moved common logic out and accounted for optional

red-tt-bear:


src/features/courses/course-controller.ts
```diff
@@ -25,7 +25,7 @@ import curriculumRepository from '../curriculum/curriculum-repository';
 import CurriculumUnitContent from '../../database/models/curriculum-unit-content';
 import CurriculumTopicContent from '../../database/models/curriculum-topic-content';
 import CurriculumWWTopicQuestion from '../../database/models/curriculum-ww-topic-question';
-import WebWorkDef, { Problem } from '@rederly/webwork-def-parser';
+import WebWorkDef, { Problem } from '@rederly/rederly-utils/lib/webwork-def-parser';
```
PanopticaRising, line 28: It seems a bit unintuitive to have to pull this out from a subdirectory of an npm package.

src/features/courses/course-controller.ts
```diff
@@ -1641,7 +1641,7 @@ class CourseController {
                             weight: parseInt(problem.value ?? '1'),
                             maxAttempts: parseInt(problem.max_attempts ?? '-1'),
                             hidden: false,
-                            optional: false,
+                            optional: problem.rederlyOptional === 'true'
```
PanopticaRising, line 1644: Personally, I like doing `true.toString()` to be consistent

PanopticaRising (approved):

src/features/courses/course-controller.ts
red-tt-bear: Yeah, I wanted to keep the exports the same for now, the default export itself is probably the most annoying part

In the future I might refactor

src/features/courses/course-controller.ts
red-tt-bear: Coming from the guy that does `++i` for efficiency 🤣

PR #495, opened 2021-05-29
Feature/red 415 course export

red-tt-bear:


src/features/courses/course-controller.ts
```diff
@@ -6072,6 +6072,65 @@ You can contact your student at ${options.student.email} or by replying to this
             }
         });
     }
+
+    fetchDataForExport({
```
PanopticaRising, line 6076: We should probably call this 'Archive' for consistency.

PanopticaRising (approved):

src/features/courses/course-controller.ts
red-tt-bear: That a good point
`CourseArchiveExport` is a good name

PR #502, opened 2021-06-10
Feature/superadmin stage

PanopticaRising:


src/features/universities/university-route-validation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const postUniversity = {
+    params: {},
+    query: {},
+    body: {
+        universityName: Joi.string().required(),
+        profEmailDomain: Joi.string().required(),
+        studentEmailDomain: Joi.string().required(),
+        verifyInstitutionalEmail: Joi.bool().optional().default(true),
+        paidUntil: Joi.date().optional().default(new Date()),
```
red-tt-bear, line 20: I don't think this does what you think it does,

This will default paidUntil to server startup date

Probably fine either way

src/features/users/user-route.ts
```diff
@@ -205,6 +208,32 @@ router.get('/status',
         next(httpResponse.Ok('fetched user status', response));
     }));
 
+    router.post('/adminUpdate',
```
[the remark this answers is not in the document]
red-tt-bear: Maybe should be `admin-user-update` otherwise it is unclear

src/features/permissions/roles.ts
```diff
@@ -2,6 +2,7 @@ enum Role {
     STUDENT = 0,
     PROFESSOR = 1,
     ADMIN = 2,
+    SUPERADMIN = 3,
```
red-tt-bear, line 5: Should probably be `SUPER_ADMIN`

src/features/users/user-route.ts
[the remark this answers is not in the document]
red-tt-bear: Oh, it's under users, that should be fine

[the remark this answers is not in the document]
red-tt-bear: Though it should probably be `super-admin-update`

src/features/users/user-route.ts
```diff
@@ -205,6 +208,32 @@ router.get('/status',
...
+            throw new Error(Constants.ErrorMessage.NIL_SESSION_MESSAGE);
+        }
+
+        const role = req.rederlyUserRole ?? req.rederlyUser?.roleId ?? Role.STUDENT;
+        if (role !== Role.SUPERADMIN) {
+            throw new ForbiddenError('You do not have access to update payment information.');
+        }
+
+        const result = await User.update({
+            paidUntil: req.body.paidUntil,
```
red-tt-bear, line 226: These are optional, what if you send it undefined for either value (since they can be undefined)?

I think this should be wrapped in `_.omitBy(updates, _.isUndefined)`

tommy-lettieri (approved):

src/features/universities/university-route-validation.ts
PanopticaRising: Good catch, I didn't think about that. It is probably still fine.

PR #505, opened 2021-06-15
feat: Added roleId to super admin update.

PanopticaRising:


src/features/users/user-route-validation.ts
```diff
@@ -134,5 +134,6 @@ export const adminUpdateValidation = {
         verified: Joi.boolean().optional(),
         firstName: Joi.string().optional(),
         lastName: Joi.string().optional(),
+        roleId: Joi.number().min(0).max(1).optional(),
```
red-tt-bear, line 137: `allow([0,1])` would have been better so someone messing around couldn't pass `1.5`

They will just get a sequelize error so it doesn't really matter

tommy-lettieri (approved):

